PR #4, opened 2024-12-16
Implement screen track capabilities

jptaylor:


Regaddi (approved):
This looks good to me!

transports/realtime-websocket-transport/package.json
```diff
@@ -21,14 +21,14 @@
     "lint": "eslint . --ext ts --report-unused-disable-directives --max-warnings 0"
   },
   "devDependencies": {
-    "@pipecat-ai/client-js": "~0.3.0",
```
mattieruth, line 24: i'm not the packaging expert, but my understanding when I went through this with @Regaddi was that all these were left at 0.3.0 because the ~ was supposed to make npm do the right thing without being so specific.

jptaylor: @mattieruth That may be true, but the screen sharing capabilities that we've added to the base transport interface will cause errors if you do not use the `0.3.2` release of the client library (where it is defined.)

It may actually be worth moving that base class over to this library, now I think about it. `client-js` only really uses it for type checking, but there are other ways we can achieve that.

mattieruth (approved):

PR #5, opened 2024-12-18
Rename RealTimeWebSocketTransport to LLMDirectTransport

mattieruth:
RealTime is too OpenAI specific and it really has nothing to do with WebSockets. This abstract Transport class is really about being a transport that interface directly with an LLM, not requiring any intermediary server.

This is a precursor to adding direct transports for openai.


mattieruth:
backing off of this. The RealTimeWebSocketTransport does essentially assume a WebSocket connection since it assumes the media is transferred via data buffers. I still think a rename would be good, but maybe to something more like `LLMDirectWebSocketTransport`

PR #9, opened 2025-01-02
Update README.md in examples/directToLLMTransports

gianpaj:
https://github.com/pipecat-ai/pipecat-client-web-transports/blob/b1ff856b49b32e5b6b16d160bbaaab206e33a43a/transports/openai-realtime-webrtc-transport/src/OpenAIRealTimeWebRTCTransport.ts#L26

### DIFF
--- a/examples/directToLLMTransports/README.md
+++ b/examples/directToLLMTransports/README.md
@@ -97,13 +97,15 @@ const llm_service_options: GeminiLLMServiceOptions = {
 };
 ```
 
-### Gemini Multimodal Live Service Configuration
+### OpenAI Multimodal Live Service Configuration
 ```typescript
 const llm_service_options: OpenAIServiceOptions = {
   api_key: import.meta.env.VITE_DANGEROUS_OPENAI_API_KEY,
   // ... additional configuration
 };
 ```
+
+By default it uses `gpt-4o-realtime-preview-2024-12-17`
 
 ## Notes
 - Gemini integration currently does not support transcripts


examples/directToLLMTransports/README.md
```diff
@@ -97,16 +97,18 @@ const llm_service_options: GeminiLLMServiceOptions = {
...
-### Gemini Multimodal Live Service Configuration
+### OpenAI Multimodal Live Service Configuration
 ```typescript
 const llm_service_options: OpenAIServiceOptions = {
   api_key: import.meta.env.VITE_DANGEROUS_OPENAI_API_KEY,
   // ... additional configuration
 };
 ```
 
+By default it uses `gpt-4o-realtime-preview-2024-12-17`
```
markbackman, line 108: This is the current default, but this is not a setting in the pipecat client web transport. For the purpose of the README, it might be better to list this as the OpenAI default model and link to OpenAI documentation.

examples/directToLLMTransports/README.md
```diff
@@ -97,16 +97,18 @@ const llm_service_options: GeminiLLMServiceOptions = {
 };
 ```
 
-### Gemini Multimodal Live Service Configuration
+### OpenAI Multimodal Live Service Configuration
```
markbackman, line 100:
```suggestion
### OpenAI Realtime API Service Configuration
```

markbackman:
Thanks @gianpaj! Just a few comments.

mattieruth (approved):
good catch!

examples/directToLLMTransports/README.md
mattieruth: doh. github links 👎 -- took me to the single commit that only had the change on 100 so I missed this. I'll submit a follow-up PR since I was the one to botch this.

gianpaj: i'm not sure where to put this but the default should be `gpt-4o-realtime-preview` so when a new version comes out you don't have update to point to the specific new version.

https://platform.openai.com/docs/models#gpt-4o-realtime

Also this, is an example README. Feel free to update as you think makes more sense :)

PR #10, opened 2025-01-13
Update example README to point to service options and defaults

mattieruth:
@gianpaj -- let me know if this serves the  purposes you were hoping to accomplish.


gianpaj:
Hi @mattieruth , that looks better, yes :)

PR #12, opened 2025-02-06
Transports improvements for the instant voice demo

filipi87:
Transports improvements for the instant voice demo


transports/daily/package.json
```diff
@@ -30,12 +30,13 @@
     "@pipecat-ai/client-js": "~0.3.2"
   },
   "dependencies": {
-    "@daily-co/daily-js": "^0.73.0"
+    "@daily-co/daily-js": "^0.73.0",
+    "@pipecat-ai/realtime-websocket-transport": "~0.3.4"
```
mattieruth, line 34: i know this might be crazy... but at this point it might make sense to pull the library of raw media management stuff into it's own library that these transports all depend on, rather than making the daily transport rely on the realtime-websocket-transport (seems weird)

transports/daily/src/transport.ts
```diff
@@ -24,34 +24,107 @@ import {
...
+  WavMediaManager,
+} from "@pipecat-ai/realtime-websocket-transport";
+
 export interface DailyTransportAuthBundle {
   room_url: string;
   token: string;
 }
 
 export interface DailyTransportConstructorOptions {
+  enableEarlyMicAccess?: boolean;
```
mattieruth, line 37: i feel like the name `earlyMicAccess` is misleading. when i read this initially, i thought "why can't we just start the mic before joining the call?" -- but it's not about starting the mic early, it's about capturing the audio from the mic and buffering/recording it so that those bytes aren't lost while waiting for the bot to join.  Maybe `bufferLocalAudio` or a more verbose `bufferLocalAudioUntilReady` with events `audio-buffering-started/stopped`?

transports/daily/src/transport.ts
```diff
@@ -24,34 +24,107 @@ import {
...
   private _dailyFactoryOptions: DailyFactoryOptions;
+  private _enableEarlyMicAccess: boolean;
   private _botId: string = "";
   private _selectedCam: MediaDeviceInfo | Record<string, never> = {};
   private _selectedMic: MediaDeviceInfo | Record<string, never> = {};
   private _selectedSpeaker: MediaDeviceInfo | Record<string, never> = {};
 
+  private static RECORDER_SAMPLE_RATE = 16_000
+  private audioQueue: ArrayBuffer[] = [];
+  private declare mediaManager: WavMediaManager
```
mattieruth, line 57: I recommend you try using the DailyMediaManager or better, extending, or modifying it so that you get proper device management. Right now you do a song-dance where you don't start the daily audio and then turn it on once the bot is ready, but i think there's a way you can forego that and just pass the audio you get from the daily track to the recorder.  Right now `DailyMediaManager` has it's own call instance, but maybe you can add a way to provide the instance?

mattieruth: (if you do a search in this repo, you'll see I've purposely NOT used the WavMediaManager because of it's minimal implementation of device management)

transports/daily/package.json
filipi87: Yeah, I agree, it does feels a little bit weird.
Should we keep this library in the same repo ?

transports/daily/src/transport.ts
filipi87: Hi @mattieruth, I had followed this route for a couple of reasons:
- `DailyMediaManager` would probably be a lot slower to initialize since we would need to download `callBundle`, wait for Daily to start the devices, and handle everything else that usually happens when creating a Daily call.
- Like you mentioned above, `DailyMediaManager` has its own call instance, and I was trying to avoiding unnecessary modifications inside the transports.
- The fact that `WavMediaManager` is a minimal implementation seemed like an advantage, as all I needed was to capture raw audio before joining a call. I even considered whether it would be possible to use only the recording instead, but I haven’t tested that approach.

Also, the reason I took this approach, enabling Daily's audio only once the bot was ready, was to avoid sending the same audio to Pipecat through WebRTC. So, while the bot was disconnected, I captured the audio using the simple `WavMediaManager`. Once the bot was ready, I enabled Daily audio, stopped the recording, and flushed the audio.

This approach has been working well in my tests so far. But maybe it would be nice a sync chat to understand what you have in mind, and if you think it would keep these requirements of low latency, and not sending the same audio twice to Pipecat.

transports/daily/src/transport.ts
filipi87: I like the idea of using `enable` to convey that we are activating something that is not enabled by default. And the reason to use `early`, was to give the sense that it is happening before we are fully ready.

But yeah, I agree that `enableEarlyMicAccess` was probably not descriptive enough.   =(

But when I think about buffer, it feels like the SDK is going to buffer, but I still need to do something with that later. 

What do you think about `enableEarlyMicCapture` instead ? In this case the events would be `EARLY_MIC_CAPTURE_STARTED` and `EARLY_MIC_CAPTURE_STOPPED`.

or `enableSpeakBeforeReady`, In this case the events would be `SPEAK_BEFORE_READY_STARTED` and `SPEAK_BEFORE_READY_STOPPED`.

transports/daily/src/transport.ts
filipi87: Another possible approach that that could be worth to discuss, it is to use directly the `MediaStreamRecorder` inside the `DailyTransport`.

transports/daily/src/transport.ts
filipi87: Decided during chat:
`bufferLocalAudioUntilBotReady` with events `audio-buffering-started/stopped`

filipi87: Done!

transports/daily/src/transport.ts
filipi87: Done, refactored to use `MediaStreamRecorder`. This approach makes everything slower in about 01 second.
cc: @markbackman

transports/daily/CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   once for each `track-started` event. Now, `clientReady()` is called for the
   first track only.
 
+- Added support for buffering audio until the bot is ready using the `bufferLocalAudioUntilBotReady`
+  property. Once the bot is ready, the buffered audio will be sent.
```
mattieruth, line 17: nit: maybe add the why

```suggestion
- Added support for buffering audio until the bot is ready using the `bufferLocalAudioUntilBotReady`
  property. Once the bot is ready, the buffered audio will be sent, allowing the user to begin speaking before the bot has joined the call.
```

transports/daily/src/transport.ts
```diff
@@ -24,34 +24,106 @@ import {
...
   private _botId: string = "";
   private _selectedCam: MediaDeviceInfo | Record<string, never> = {};
   private _selectedMic: MediaDeviceInfo | Record<string, never> = {};
   private _selectedSpeaker: MediaDeviceInfo | Record<string, never> = {};
 
+  private static RECORDER_SAMPLE_RATE = 16_000
+  private static RECORDER_CHUNK_SIZE = 512
+  private currentAudioTrack: MediaStreamTrack | null = null;
+  private audioQueue: ArrayBuffer[] = [];
+  private declare mediaStreamRecorder: MediaStreamRecorder;
```
mattieruth, line 57: nit: perhaps stick with the `_` prefix for private variables?

transports/daily/src/transport.ts
```diff
@@ -24,34 +24,106 @@ import {
...
     dailyFactoryOptions = {},
+    bufferLocalAudioUntilBotReady = false,
   }: DailyTransportConstructorOptions = {}) {
     super();
     this._dailyFactoryOptions = dailyFactoryOptions;
+    this._bufferLocalAudioUntilBotReady = bufferLocalAudioUntilBotReady;
+  }
+
+  private setupRecorder(): void {
+    this.mediaStreamRecorder = new MediaStreamRecorder({ sampleRate: DailyTransport.RECORDER_SAMPLE_RATE });
```
mattieruth, line 69: wrap at 80? do we have a linter/prettier set up for our pipecat js stuffs?

transports/daily/src/transport.ts
```diff
@@ -274,6 +367,17 @@ export class DailyTransport extends Transport {
     });
   }
 
+  private stopRecording() {
+    if (this._bufferLocalAudioUntilBotReady) {
```
mattieruth, line 371: should this look more like the check in `disconnect` so that we don't have someone flip this flag and then the recorder keeps going?

```suggestion
    if (this.mediaStreamRecorder && this.mediaStreamRecorder.getStatus() !== "ended")
```

transports/daily/src/transport.ts
```diff
@@ -305,6 +409,12 @@ export class DailyTransport extends Transport {
     this._daily.stopLocalAudioLevelObserver();
     this._daily.stopRemoteParticipantsAudioLevelObserver();
 
+    this.audioQueue = []
+    this.currentAudioTrack = null
+    if (this.mediaStreamRecorder && this.mediaStreamRecorder.getStatus() !== "ended") {
+      void this.mediaStreamRecorder.end()
+    }
```
mattieruth, line 416: could this be replaced with a call to `stopRecording()`? Is there a reason we wouldn't want to send the recording-stopped event in this case?

mattieruth (approved):
looking good! mostly nits and question 👍  approving so i'm not blocking

transports/daily/src/transport.ts
filipi87: Yep, it makes sense.

transports/daily/src/transport.ts
filipi87: done

transports/daily/src/transport.ts
filipi87: Done.

transports/daily/src/transport.ts
filipi87: Done.

transports/daily/src/transport.ts
filipi87: Done.

transports/daily/src/transport.ts
```diff
@@ -355,6 +465,35 @@ export class DailyTransport extends Transport {
...
+        break;
+      case "paused":
+        await this.startRecording();
+        break;
+      case "recording":
+      default:
+        if (this.currentAudioTrack !== track) {
+          await this.mediaStreamRecorder.end();
+          await this.mediaStreamRecorder.begin(track);
+          await this.startRecording();
```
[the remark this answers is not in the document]
filipi87: Yeah, I have just tried that, and no errors, everything keeps working again if we unmute. So I think we are good to go. =)

PR #14, opened 2025-02-27
Create an accessible proxy around the Daily call instance that works …

mattieruth:
…just like the Daily instance but rejects calls that changes the call's lifecycle

This is a draft of one approach for making the internal daily instance accessible and usable so users can access all the features daily offers.

The primary concern I see with simply making the instance public is that calling things like `join()` would bypass the RTVIClient's process for connecting and get state out-of-whack. maybe we eventually support this but for now, an easy way to avoid this is simply to proxy the call and reject calls to things that modify the call's lifecycle (join/leave/startCamera/destroy).

I haven't tested this in the wild, but thought it worth putting out for others feedback.

### DIFF
--- a/transports/daily/src/transport.ts
+++ b/transports/daily/src/transport.ts
@@ -50,14 +50,29 @@ class DailyCallWrapper {
     this._proxy = new Proxy(this._daily, {
       get: (target, prop, receiver) => {
         if (typeof target[prop as keyof DailyCall] === "function") {
-          // Disable certain methods
-          if (
-            ["preAuth", "startCamera", "join", "leave", "destroy"].includes(
-              String(prop)
-            )
-          ) {
+          let errMsg;
+          switch (String(prop)) {
+            // Disable methods that modify the lifecycle of the call. These operations
+            // should be performed via the RTVI client in order to keep state in sync.
+            case "preAuth":
+              errMsg = `Calls to preAuth() are disabled.`;
+              break;
+            case "startCamera":
+              errMsg = `Calls to startCamera() are disabled. Please use RTVIClient.initDevices()`;
+              break;
+            case "join":
+              errMsg = `Calls to join() are disabled. Please use RTVIClient.connect()`;
+              break;
+            case "leave":
+              errMsg = `Calls to leave() are disabled. Please use RTVIClient.disconnect()`;
+              break;
+            case "destroy":
+              errMsg = `Calls to destroy() are disabled.`;
+              break;
+          }
+          if (errMsg) {
             return () => {
-              throw new Error(`The ${String(prop)} method is disabled.`);
+              throw new Error(errMsg);
             };
           }
           // Forward other method calls
@@ -73,10 +88,6 @@ class DailyCallWrapper {
 
   get proxy(): DailyCall {
     return this._proxy;
-  }
-
-  get instance(): DailyCall {
-    return this._daily;
   }
 }
 
@@ -191,7 +202,7 @@ export class DailyTransport extends Transport {
     logger.debug("[RTVI Transport] Initialized");
   }
 
-  get dailyInstance(): DailyCall {
+  get dailyCallClient(): DailyCall {
     return this._dailyWrapper.proxy;
   }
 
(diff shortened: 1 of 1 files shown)


mattieruth:
also. apparently we need to add a prettier config so we aren't all linter fighting

transports/daily/src/transport.ts
```diff
@@ -150,6 +191,10 @@ export class DailyTransport extends Transport {
     logger.debug("[RTVI Transport] Initialized");
   }
 
+  get dailyInstance(): DailyCall {
```
mattieruth, line 194: maybe this should be called something like `callInstance()` so it's generic and could be a base `Transport` function that all transports should implement and could be exposed in the `Client`?

kompfner: i'm good with keeping it Daily-specific for now—it's not clear to me that other transports will even necessarily be built around some single "instance" of a client

kompfner, line 194: nit: in our documentation this is the "call client", not the "call instance"

transports/daily/src/transport.ts
```diff
@@ -41,7 +41,47 @@ export enum DailyRTVIMessageType {
...
+        return Reflect.get(target, prop, receiver);
+      },
+    });
+  }
+
+  get proxy(): DailyCall {
+    return this._proxy;
+  }
+
+  get instance(): DailyCall {
```
kompfner, line 78: this isn't used, right? is this just here to show how we would expose this if someday we wanted to give people "direct" (un-proxied) access to the call client?

transports/daily/src/transport.ts
```diff
@@ -41,7 +41,47 @@ export enum DailyRTVIMessageType {
...
+class DailyCallWrapper {
+  private _daily: DailyCall;
+  private _proxy: DailyCall;
+
+  constructor(daily: DailyCall) {
+    this._daily = daily;
+    this._proxy = new Proxy(this._daily, {
+      get: (target, prop, receiver) => {
+        if (typeof target[prop as keyof DailyCall] === "function") {
+          // Disable certain methods
```
kompfner, line 53: nit: comment here should explain why

transports/daily/src/transport.ts
```diff
@@ -41,7 +41,47 @@ export enum DailyRTVIMessageType {
...
+          if (
+            ["preAuth", "startCamera", "join", "leave", "destroy"].includes(
+              String(prop)
+            )
+          ) {
+            return () => {
+              throw new Error(`The ${String(prop)} method is disabled.`);
```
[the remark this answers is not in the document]
mattieruth: yeah, my plan is actually to expand each one out to say what they should use instead  (i.e. `connect`/`initDevices`/`disconnect`). i just didn't want to spend cycles yet cleaning everything up if this wasn't the right approach. but now... full steam ahead!

transports/daily/src/transport.ts
mattieruth: right. it just felt like it rounded out the implementation.

transports/daily/src/transport.ts
mattieruth: good point. 👍

mattieruth: oh. that comment was meant for the next comment about "call client". as for others... i mean, i could definitely see someone eventually contributing a twilio or livekit transport. but i'll leave as-is for now. if you are doing vendor-specific things, you can call a vender-specific method to get the underlying instance.

transports/daily/src/transport.ts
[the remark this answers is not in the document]
kompfner: > expand each one out to say what they should use instead

great idea!

mattieruth:
I have updated this PR and addressed the feedback. I also tested the changes with the simple chatbot example and it works as expected. The one question I have is that I made `dailyCallClient` a `get`er so using it looks like:

`myTransport.dailyCallClient.startRecording()`

Do we want the getter to be function instead? I'm indifferent.

transports/daily/src/transport.ts
kompfner: If it were up to me I'd get rid of this since it's just dead code; I prefer to add code when it's needed rather than "just in case" we may want it someday. But don't consider this feedback blocking if you just want to merge. I'm good either way.

kompfner (approved):
LGTM!

PR #15, opened 2025-03-13
Support preAuth

mattieruth:
From the first commit... This PR

Updates the daily instance in the daily transport to:
1. Make it available immediately upon Transport construction
2. Fix use of daily factory options to work better alongside rtvi options
3. Add support for calling preAuth()
4. Change how disconnect works such that the transport no longer re-creates
   a new daily instance but re-uses the one it has so there is 1:1
   transport -> daily instance.
   
And updates the OpenAIRealTimeWebRTCTransport to work with the changes laid out in this [client-js PR](https://github.com/pipecat-ai/pipecat-client-web/pull/98)

### DIFF
--- a/transports/daily/src/transport.ts
+++ b/transports/daily/src/transport.ts
@@ -16,6 +16,7 @@ import {
   Participant,
   RTVIClientOptions,
   RTVIError,
+  RTVIEventCallbacks,
   RTVIMessage,
   Tracks,
   Transport,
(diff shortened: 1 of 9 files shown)


kompfner (approved):

transports/daily/src/transport.ts
```diff
@@ -475,7 +495,6 @@ export class DailyTransport extends Transport {
     this.stopRecording();
 
     await this._daily.leave();
-    await this._daily.destroy();
```
filipi87, line 478: I thought we had a release or destroy method where we could include this, but I was wrong. 🤕

filipi87 (approved):
Looks good. Should we already mention this in the [changelog](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/daily/CHANGELOG.md) ?

mattieruth:
> Looks good. Should we already mention this in the [changelog](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/daily/CHANGELOG.md) ?

done. Also just added a commit to bump the client-js dependencies to the latest release (which this depends on)

PR #16, opened 2025-03-13
Bump transport versions in preparation for release

mattieruth:


transports/daily/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to **Pipecat Daily WebRTC Transport** will be documented in
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.3.5]
```
filipi87, line 8: Just looking and that is probably on me, but the 0.3.5 has already been released last month (`2025-02-13`). 

I guess with the features to support the buffering audio.

filipi87: https://www.npmjs.com/package/@pipecat-ai/daily-transport

transports/gemini-live-websocket-transport/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipecat-ai/gemini-live-websocket-transport",
-  "version": "0.3.4",
+  "version": "0.3.5",
```
filipi87, line 3: Should we go ahead and bump this directly to `0.3.6` to keep all the transports in sync ? What do you think ?

filipi87: Actually, never mind, there is probably no reason for all of them to have the same version.

filipi87 (approved):
I guess we just need to fix the changelog to include the release date of version 0.3.5 and also add the release date for the new version 0.3.6.

transports/daily/CHANGELOG.md
mattieruth: doh! good catch. fixed.

PR #26, opened 2025-03-31
call initDevices from initialize

Regaddi:
This makes sure that media devices are initialized correctly.


transports/openai-realtime-webrtc-transport/src/OpenAIRealTimeWebRTCTransport.ts
```diff
@@ -143,6 +143,8 @@ export class OpenAIRealTimeWebRTCTransport extends Transport {
       this._attachDeviceListeners();
     }
 
+    this.initDevices();
```
mattieruth, line 146: i don't think we want this, but i'm curious the impetus for adding it. this `initialize()` gets called from the `RTVIClient`'s constructor and feels like a weird action to turn on your devices. Right now, we've been working on the theory that if a user wants to initialize devices before `connect()` they can manually call `initDevices()`, otherwise, they automatically turn on/init on `connect()`.

PR #31, opened 2025-04-08
Sending a signalling message each time a track is enabled or disabled.

filipi87:


transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -9,18 +9,38 @@ import {
...
+  type = "trackStatus";
+  receiver_index: number;
+  enabled: boolean;
+  constructor(receiver_index: number, enabled: boolean) {
+    this.receiver_index = receiver_index;
+    this.enabled = enabled;
+  }
 }
 
+const RENEGOTIATE_TYPE = "renegotiate" as const;
```
mattieruth, line 22: curious, why the `as const`?

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -9,18 +9,38 @@ import {
...
+type RenegotiateMessage = {
+  type: typeof RENEGOTIATE_TYPE;
+};
```
mattieruth, line 25: also, why is this a type and type of vs following the pattern of TrackStatusMessage?

```suggestion
class RenegotiateMessage = {
  type = RENEGOTIATE_TYPE;
};
```

mattieruth (approved):
nit comments/questions, but generally LGTM!

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
filipi87: We don't actually need it, probably just a bad code completion from my IDE.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
filipi87: Fixed.

PR #33, opened 2025-04-15
Fix: Add 500ms delay to client-ready message for iOS devices

markbackman:
I tested this locally numerous times. For the majority of cases, the track audio starts to play within 250ms, but there are enough occurrences where audio can take up to 500ms that we need to use 500ms as a delay.

### DIFF
--- a/transports/daily/src/transport.ts
+++ b/transports/daily/src/transport.ts
@@ -429,19 +429,45 @@ export class DailyTransport extends Transport {
 
   async sendReadyMessage(): Promise<void> {
     return new Promise<void>((resolve) => {
-      (async () => {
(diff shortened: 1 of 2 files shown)


transports/daily/src/transport.ts
```diff
@@ -429,19 +429,45 @@ export class DailyTransport extends Transport {
...
+        resolve();
+      };
+
+      const readyHandler = (ev: DailyEventObjectTrack) => {
+        if (!ev.participant?.local) {
+          this._daily.off("track-started", readyHandler);
+
+          // Check if it's an iOS device
+          if (isIOS()) {
+            logger.info(
```
mattieruth, line 455: nit: debug log instead?

mattieruth (approved):

PR #36, opened 2025-04-22
refactor(webrtc): Improve ICE server configuration in SmallWebRTCConnection

ken-kuro:
This PR refactors the `SmallWebRTCConnection` constructor, introduces a new constructor option, `waitForICEGathering` (boolean, defaults to `false`) to improve how ICE servers are handled.

**Changes:**

* Added `waitForICEGathering` property to newly added `SmallWebRTCTransportConstructorOptions` interface and update the stored private member variable `_waitForICEGathering` within the `SmallWebRTCTransport` class via the constructor.
* Updated the `negotiate` method to use the stored instance configuration to control the waiting logic.
* The `_iceServers` private member variable in `SmallWebRTCTransport` class now accepts `RTCIceServer[]` instead of just the `string[]` of urls.
* Update the setter of `iceServers` for the new type change.

**Motivation:**

* This change allows users to pass pre-configured `RTCIceServer` objects directly, which is particularly useful for more complex setups involving TURN servers that require credentials.
* This allows users to configure whether the transport should explicitly wait for the `iceGatheringState` to become `complete` during the negotiation phase (`negotiate` method) before sending the SDP offer. It would provides more control over the WebRTC connection setup, particularly useful in scenarios where ensuring all ICE candidates are gathered is necessary before proceeding, or conversely, when faster setup is preferred by not waiting.
*  It improves type safety and makes the expected input clearer.

### DIFF
--- a/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
+++ b/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
@@ -19,6 +19,10 @@ class TrackStatusMessage {
   }
 }
 
+export interface SmallWebRTCTransportConstructorOptions {
+  waitForICEGathering?: boolean;
+}
+
 const RENEGOTIATE_TYPE = "renegotiate";
 class RenegotiateMessage {
   type = RENEGOTIATE_TYPE;
@@ -73,9 +77,13 @@ export class SmallWebRTCTransport extends Transport {
   private keepAliveInterval: number | null = null;
 
   private _iceServers: RTCIceServer[] = [];
+  private readonly _waitForICEGathering: boolean
 
-  constructor() {
+  constructor({
+    waitForICEGathering = false
+  }: SmallWebRTCTransportConstructorOptions = {}) {
     super();
+    this._waitForICEGathering = waitForICEGathering;
     this.mediaManager = new DailyMediaManager(
       false,
       false,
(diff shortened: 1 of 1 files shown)


mattieruth (commented):
Thanks for the PR! I've tested the changes and in general approve. I'm going to leave it to @filipi87  for final approval, though. The `_iceServers` change would be a breaking change and I'm not sure if he had reasons for typing just the `urls`.  Maybe for convenience?

ken-kuro:
I see, thank you. Have a great day!

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -19,6 +19,10 @@ class TrackStatusMessage {
   }
 }
 
+export interface SmallWebRTCTransportConstructorOptions {
```
filipi87, line 22: It would be nice if we could also receive the ice servers here.

filipi87 (approved):
The changes make perfect sense. This refactor was actually on my todo list. 🙂
Thanks for taking care of it, @ken-kuro.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
ken-kuro: Got it, I will update it right away

PR #38, opened 2025-05-12
small-webrtc-transport: set state "ready"

vipyne:
This makes RTVI actions work with small webrtc transport.

Unless [transport state is "ready"](https://github.com/pipecat-ai/pipecat-client-web/blob/ab3fc9801f4d0be16b68be62e1aca6abdf432729/client-js/src/decorators.ts#L18), RTVI actions [will not work](https://github.com/pipecat-ai/pipecat-client-web/blob/ab3fc9801f4d0be16b68be62e1aca6abdf432729/client-js/src/decorators.ts#L22).


filipi87 (approved):
Yep, nice catch. 🚀

PR #42, opened 2025-05-26
Creating a Pipecat websocket transport.

filipi87:
- Renamed `RealTimeWebsocketTransport` to `DirectToLLMTransport`.
- Created a `WebSocketTransport` to connect with Pipecat.


filipi87:
I went back and forth on whether this should be a separate npm module.
But when I see the name: `@pipecat-ai/realtime-websocket-transport`, it feels like the right module to install.

So for now, I’ve added `PipecatWebsocketTransport` here, in this same module. No hard feelings if you think it should live in its own npm package. I’m happy to move it.

That said, I’m not entirely happy with the name `PipecatWebsocketTransport`.

Our other transports are named like `DailyTransport`, `SmallWebRTCTransport` ..., without mentioning Pipecat.
But here, we already expose the abstract class `RealtimeWebsocketTransport`, which is also used by `GeminiLiveWebsocketTransport`.

So calling this one just `WebsocketTransport` might cause confusion. For example, how would someone know when to use `WebsocketTransport` vs `RealtimeWebsocketTransport` ?

One alternative could be to rename the abstract class `RealtimeWebsocketTransport` to something else. But this would be a breaking change.

Totally open to suggestions here.

lib/media-mgmt/dailyMediaManager.ts
```diff
@@ -31,11 +31,15 @@ export class DailyMediaManager extends MediaManager {
...
 
+  private _recorderChunkSize: number | undefined = undefined;
+
   constructor(
     enablePlayer: boolean = true,
     enableRecording: boolean = true,
     onTrackStartedCallback?: (event: DailyEventObjectTrack) => void,
     onTrackStoppedCallback?: (event: DailyEventObjectTrack) => void,
+    recorderChunkSize: number | undefined = undefined,
+    recorderSampleRate: number | undefined = 24000,
```
mattieruth, line 42: if this defaults to 24000, should it support undefined?

transports/realtime-websocket-transport/src/pipecatWebSocketTransport.ts
```diff
@@ -0,0 +1,225 @@
...
+      }
+    });
+    this._ws.on("error", (error: Error) => {
+      this.connectionError(`websocket error: ${error}`);
+    });
+    this._ws.on("connection-timeout", () => {
+      this.connectionError("websocket connection timed out");
+    });
+    this._ws.on("close", (code: number) => {
+      this.connectionError(`websocket connection closed. Code: ${code}`);
```
mattieruth, line 87: this is triggering an error, but is close always an error? what about on actual disconnect?

mattieruth (commented):
I'm torn on telling you not to extend from `RealTimeWebsocketTransport` and instead just be your own Transport. I get why you did it, so that you could benefit from the existing mediaManager bits but [as the documentation says](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/realtime-websocket-transport/src/realTimeWebSocketTransport.ts#L20-L24),  it was really designed with the purpose in mind of connecting directly to an LLM and not a pipecat server. It's why LLMServiceOptions are integral and require an API key and why there are functions to initialize and setup the LLM.  

so i don't know. it's a weird use of it, but i also get it. what do you think?

mattieruth:
as for naming... Would renaming `RealtimeWebsocketTransport` really be a breaking change since it's an abstract class and should only really be used internally by the Gemini and OpenAI transports?

If not and if you change the new transport not to extend from it, maybe we rename it to better represent what I said. Something like `DirectToLLMTransport` 🤷‍♀️

transports/realtime-websocket-transport/src/pipecatWebSocketTransport.ts
```diff
@@ -0,0 +1,225 @@
...
+import { ReconnectingWebSocket } from "../../../lib/websocket-utils/reconnectingWebSocket.js";
+import { RealTimeWebsocketTransport } from "./realTimeWebSocketTransport.ts";
+import { DailyMediaManager } from "../../../lib/media-mgmt/dailyMediaManager.ts";
+
+import { Frame } from "./generated/proto/frames.ts";
+
+export class PipecatWebsocketTransport extends RealTimeWebsocketTransport {
+  declare private _ws: ReconnectingWebSocket | null;
+  declare private _botResponseID: number;
+  private static RECORDER_SAMPLE_RATE = 16_000;
```
marcus-daily, line 16: This defaults to 24000Hz in DailyMediaManager -- is this intentional?

transports/realtime-websocket-transport/src/pipecatWebSocketTransport.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  attachLLMListeners(): void {
+    // nothing to do here
+  }
+
+  initializeWebsocket(authBundle: any): void {
+    this._ws = new ReconnectingWebSocket(`${authBundle.ws_url}`, undefined, {
+      parseBlobToJson: false,
+    });
+    // disabling the keep alive, there is no API for it inside Pipecat
```
marcus-daily, line 43: I'm not sure what this comment means -- do we need an API for keepalive?

transports/realtime-websocket-transport/proto/frames.proto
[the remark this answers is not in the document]
filipi87: Done.

transports/realtime-websocket-transport/src/pipecatWebSocketTransport.ts
filipi87: Yes, it is. 
DailyMediaManager is already used by other transports, like Gemini and OpenAi, for direct connecting with the LLM, and in those cases they expect to receive 24000Hz, which was the default value.

But for Pipecat, the default that it expects, and in order to work with the SileroVad it is 16000Hz.

transports/realtime-websocket-transport/src/pipecatWebSocketTransport.ts
filipi87: The ReconnectingWebSocket is already used by other transports which support keep alive. 
We don't have that yet inside Pipecat. 
Maybe in the future we can start to support that.

transports/realtime-websocket-transport/src/pipecatWebSocketTransport.ts
```diff
@@ -0,0 +1,225 @@
...
+    this._ws.on("open", () => {
+      logger.debug("Websocket connection opened");
+    });
+    this._ws.on("message", async (data: Blob) => {
+      let arrayBuffer: ArrayBuffer = await data.arrayBuffer();
+      const parsedFrame = Frame.fromBinary(new Uint8Array(arrayBuffer)).frame;
+      if (parsedFrame.oneofKind === "audio") {
+        if (!this._botIsSpeaking) {
+          this._botResponseID++;
+          this.botStartedSpeaking();
```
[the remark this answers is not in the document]
filipi87: Yep, this was not needed. Nice catch.
I have removed it. 🙂

mattieruth (approved):
two small questions but i approve either way :). nice work! and thanks for the cleanup!

PR #44, opened 2025-06-04
Turning lib into a internal package

filipi87:
Turning lib into a internal package

### DIFF
--- a/transports/realtime-websocket-transport/src/index.ts
+++ b/transports/realtime-websocket-transport/src/index.ts
@@ -1,8 +1,1 @@
-// export * from "./realTimeWebSocketTransport";
-// export * from "../../../lib/wavtools/dist/index.d.ts";
-
-import { WavMediaManager } from "../../../lib/media-mgmt/mediaManager";
-import { DailyMediaManager } from "../../../lib/media-mgmt/dailyMediaManager";
-import { WebSocketTransport } from "./webSocketTransport.ts";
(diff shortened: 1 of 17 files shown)


transports/realtime-websocket-transport/src/index.ts
```diff
@@ -1,8 +1 @@
-// export * from "./realTimeWebSocketTransport";
-// export * from "../../../lib/wavtools/dist/index.d.ts";
-
-import { WavMediaManager } from "../../../lib/media-mgmt/mediaManager";
-import { DailyMediaManager } from "../../../lib/media-mgmt/dailyMediaManager";
-import { WebSocketTransport } from "./webSocketTransport.ts";
-
-export { WavMediaManager, DailyMediaManager, WebSocketTransport };
+export * from "./webSocketTransport";
```
mattieruth, line 1: the fact that the folder is realtime-websocket-transport but the transport is just websockettransport is still throwing me.

mattieruth (approved):
makes sense to me.

filipi87:
I thought this approach was working, but when I create the npm package (using `npm pack`) and try to use the actual package, which is the same one we would publish, it complains about not finding the internal package: `@pipecat-ai/web-transports-lib`.

So, the only way to solve it, is if we actually published that package.

I am noving this PR for draft until we decide the best way to do this.

PR #50, opened 2025-07-25
Add support for overriding the mediaManager

mattieruth:
while we highly discourage using the WavMediaManager since Daily's media management is immensely more feature-complete, we shouldn't make it impossible.


filipi87 (approved):
Makes a lot of sense. Nice improvement.

PR #51, opened 2025-08-01
Implement DeviceError handling in transports

mattieruth:
This PR goes hand-in-hand with https://github.com/pipecat-ai/pipecat-client-web/pull/134, which introduces a new DeviceError callback. This PR implements the transport portions for actually trigging those callbacks.

### DIFF
--- a/transports/daily/src/transport.ts
+++ b/transports/daily/src/transport.ts
@@ -649,13 +649,17 @@ export class DailyTransport extends Transport {
         }
         case "cam-in-use": {
           devices.push("cam");
-          return new DeviceError(devices, error.type, error.msg);
+          return new DeviceError(devices, "in-use", error.msg);
         }
         case "mic-in-use": {
           devices.push("mic");
-          return new DeviceError(devices, error.type, error.msg);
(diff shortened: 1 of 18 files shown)


transports/daily/src/transport.ts
```diff
@@ -608,25 +619,82 @@ export class DailyTransport extends Transport {
...
     const status = this._mediaStreamRecorder.getStatus();
     switch (status) {
       case "ended":
-        await this._mediaStreamRecorder.begin(track);
-        await this.startRecording();
+        try {
+          await this._mediaStreamRecorder.begin(track);
+          await this.startRecording();
+        } catch (e) {
+          // void. nothing to do.
```
filipi87, line 682: Same question as before, but now I'm wondering, is the reason you think we don't need to do anything here because the recorder would internally trigger a device error?

mattieruth: correct. the idea is that begin() triggers the callback and logs warnings/errors already.

filipi87 (approved):
LGTM!

PR #54, opened 2025-08-07
Update connection url naming

mattieruth:
to avoid confusion with `endpoint` and better align with other transports' naming for the url to communicate across (DailyTransport: `room_url`, and WebsocketTransport: `ws_url`), this deprecates the `connectionUrl` field name to promote a replacement of `webrtc_url`

**UPDATE**:

To be consistent with camelCasing, this is updating `ws_url` to be `wsUrl` and using `webrtcUrl` instead of `webrtc_url`. The validation step, however, is fixed so that either can be passed in in order to support existing examples and servers that really want to do everything in snake_case.

ALSO: Since we're touching the various validation methods, this updates Daily's validation to accept the current PipecatClout behavior of returning `dailyRoom`/`dailyToken`, thought that _should_ be updated in time to return simply `url`/`token`.

### DIFF
--- a/set-client-js.sh
+++ b/set-client-js.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+# Script to set client-js package across project directories
+# Usage: ./set-client-js.sh <package-spec>
+# Examples:
+#   ./set-client-js.sh /path/to/local/client-js     # Install from local path
+#   ./set-client-js.sh @pipecat-ai/client-js        # Install from npm
+#   ./set-client-js.sh @pipecat-ai/client-js@1.2.3  # Install specific version
+
+set -e  # Exit on any error
+
+# Check if package spec argument is provided
+if [ $# -eq 0 ]; then
+    echo "Error: Please provide a package specification"
+    echo "Usage: $0 <package-spec>"
+    echo ""
+    echo "Examples:"
+    echo "  $0 /path/to/local/client-js        # Install from local path"
+    echo "  $0 1.2.3                           # Install specific version"
+    exit 1
+fi
+
+PACKAGE_SPEC="$1"
+
+# Determine if this is a local path or npm package
+if [ -d "$PACKAGE_SPEC" ]; then
+    # It's a local directory
+    PACKAGE_SPEC=$(realpath "$PACKAGE_SPEC")
+    INSTALL_SPEC="@pipecat-ai/client-js@file:$PACKAGE_SPEC"
+    LOCAL_INSTALL=true
+    echo "Installing local client-js from: $PACKAGE_SPEC"
+else
+    # It's an npm package spec
+    INSTALL_SPEC="$PACKAGE_SPEC"
+    LOCAL_INSTALL=false
+    echo "Installing client-js package: $PACKAGE_SPEC"
+fi
+
+# Function to install in a directory
+install_in_dir() {
+    local dir="$1"
+    if [ -d "$dir" ] && [ -f "$dir/package.json" ]; then
+        echo "Installing in: $dir"
+        cd "$dir"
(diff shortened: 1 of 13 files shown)


set-client-js.sh
```diff
@@ -0,0 +1,85 @@
...
+            # Use local path installation
+            npm install "$INSTALL_SPEC"
+        else
+            if [[ "$dir" == *"transports"* ]]; then
+                # Install as peer and dev dependency in transport directories
+                npm install --save-peer "@pipecat-ai/client-js@~$INSTALL_SPEC"
+                npm install --save-dev "@pipecat-ai/client-js@^$INSTALL_SPEC"
+            else
+                # Use npm package installation
+                npm install "@pipecat-ai/client-js@^$INSTALL_SPEC"
```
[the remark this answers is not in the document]
mattieruth: no. i purposefully changed the script to just take a version like '1.1.0' and NOT the full path

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -24,14 +24,18 @@ class TrackStatusMessage {
 export interface SmallWebRTCTransportConstructorOptions {
   iceServers?: RTCIceServer[];
   waitForICEGathering?: boolean;
+  /** @deprecated Use webrtc_url instead */
   connectionUrl?: string;
+  webrtc_url?: string;
```
kompfner, line 29: Hmm, this ought to be camelCase, since we're in JS here (and all the other options next to it are camelCase). As is, this kinda sticks out like a sore thumb.

I'm guessing the reason for snake_case here is because the equivalent options in the other transports already are snake_case? If so...I might lobby for updating them across the board rather than doubling down on this pattern.

(I know this feels annoyingly nitpicky but IMO this kind of public API polish does make stuff feel nicer to use)

mattieruth: it's more than that. It's because these values are often sent to us from a server endpoint which has generally send responses snake cased :/

So it's more about what we want to standardize in those payloads to make `client.startBot().then((response) => client.connect(response))` just work.

might be worth a quick chat??

mattieruth: also related: https://pluot.slack.com/archives/C07V2LX3GER/p1754308816020919

set-client-js.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+# Script to set client-js package across project directories
+# Usage: ./set-client-js.sh <package-spec>
+# Examples:
+#   ./set-client-js.sh /path/to/local/client-js     # Install from local path
+#   ./set-client-js.sh @pipecat-ai/client-js        # Install from npm
+#   ./set-client-js.sh @pipecat-ai/client-js@1.2.3  # Install specific version
```
kompfner, line 8: Is this to facilitate supporting older versions of client-js?

If so—can we simplify our version matrix and just make it so that whenever we change a transport we make it require the latest client-js? Or do something to make it so that we're not maintaining a complicated version matrix?

mattieruth: no. this is because developing this library and building against a local client-js is a b*^&@ and i wanted a handy script to swap out my version for testing.  it will also make it easy to quickly update the client-js version whenever we perform that step.

mattieruth: *i wanted a handy script to swap out my local version for testing and then swap back the version that was there before accidentally committing a bunch of local paths to all the package* files. and then having to do nasty commit --amends and rebases to undo it 😂

kompfner (approved):
🥳

PR #59, opened 2025-08-26
iss-58: Support custom data and headers in webrtc connection endpoint

### LINKED ISSUE #58: SmallWebRTCTransport: Support passing header and requestParams to api/offer endpoint
It's very likely you may need to pass an auth token or other data as part of the call to your offer endpoint. Currently, there is no way to do this. We have the `startBot()` method which supports such things, but that is a separate endpoint from the `SmallWebRTCTransport`'s offer/answer endpoint. In fact, `startBot()` may be useful to generate and return the token you then need to pass to the offer endpoint and today, that's not possible.  Let's make it possible :)

mattieruth:
This change deprecates  the `webrtcUrl` field, replacing it with a new `webrtcRequestParams` field, adding support for passing a complete `APIRequest` type to define the connection endpoint.  This allows clients to pass custom header and data along to the offer/answer endpoint at connection time.  This change means that clients will want to update their code from:
```
pcClient.connect({ webrtcUrl: '/api/offer' });
```
to
```
pcClient.connect({webrtcRequestParams: { endpoint: 'api/offer' }})
```

But this allows now for things like:

```
pcClient.connect({
    webrtcUrl: {
        endpoint: '/api/offer',
        requestData: { foo: 'bar' }, // data passed in here will be passed to the server under the `requestData` field
        headers: { 'Authorization': MY_TOKEN } // headers passed in here are added to `"Content-Type": "application/json"`
}})
```

### DIFF
--- a/transports/small-webrtc-transport/package.json
+++ b/transports/small-webrtc-transport/package.json
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js",
+    "@types/lodash": "^4.17.20",
     "@types/node": "^22.9.0",
     "eslint": "9.11.1",
     "eslint-config-prettier": "^9.1.0",
@@ -33,7 +34,8 @@
   },
   "dependencies": {
     "@daily-co/daily-js": "^0.83.1",
-    "dequal": "^2.0.3"
+    "dequal": "^2.0.3",
+    "lodash": "^4.17.21"
   },
   "description": "Pipecat Small WebRTC Transport Package",
   "author": "Daily.co",
--- a/transports/small-webrtc-transport/package.json
+++ b/transports/small-webrtc-transport/package.json
@@ -22,7 +22,7 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js",
+    "@pipecat-ai/client-js": "^1.3.0",
     "@types/lodash": "^4.17.20",
     "@types/node": "^22.9.0",
     "eslint": "9.11.1",
@@ -30,7 +30,7 @@
     "eslint-plugin-simple-import-sort": "^12.1.1"
   },
   "peerDependencies": {
-    "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js"
+    "@pipecat-ai/client-js": "~1.3.0"
   },
   "dependencies": {
     "@daily-co/daily-js": "^0.83.1",
--- a/transports/small-webrtc-transport/package.json
+++ b/transports/small-webrtc-transport/package.json
@@ -22,14 +22,14 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "@pipecat-ai/client-js": "^1.2.0",
+    "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js",
     "@types/node": "^22.9.0",
     "eslint": "9.11.1",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-simple-import-sort": "^12.1.1"
   },
   "peerDependencies": {
-    "@pipecat-ai/client-js": "~1.2.0"
+    "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js"
   },
   "dependencies": {
     "@daily-co/daily-js": "^0.83.1",
--- a/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
+++ b/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
@@ -8,7 +8,9 @@ import {
   TransportStartError,
   TransportState,
   UnsupportedFeatureError,
+  APIRequest,
 } from "@pipecat-ai/client-js";
+import { makeRequest } from "@pipecat-ai/client-js";
 import { MediaManager } from "../../../lib/media-mgmt/mediaManager";
 import { DailyMediaManager } from "../../../lib/media-mgmt/dailyMediaManager";
 
@@ -37,7 +39,7 @@ export interface SmallWebRTCTransportConstructorOptions {
   waitForICEGathering?: boolean;
   /** @deprecated Use webrtcUrl instead */
   connectionUrl?: string;
-  webrtcUrl?: string;
+  webrtcUrl?: string | APIRequest;
   audioCodec?: string;
   videoCodec?: string;
   mediaManager?: MediaManager;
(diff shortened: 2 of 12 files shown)


transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -221,11 +230,20 @@ export class SmallWebRTCTransport extends Transport {
...
-    if (!this._webrtcUrl) {
+      this._offerRequest;
+    if (connectParams?.connectionUrl) {
+      logger.warn("connectionUrl is deprecated, use webrtcUrl instead");
+    }
+    if (_webrtcUrl) {
+      if (typeof _webrtcUrl === "string") {
+        this._offerRequest = { endpoint: _webrtcUrl };
+      } else {
+        this._offerRequest = _webrtcUrl;
```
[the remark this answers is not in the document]
mattieruth: i think that's the point and ok. I don't want to support merging a previously set _offerRequest with a new one. But I do like the idea of making sure it's a valid APIRequest type

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -114,12 +121,40 @@ export class SmallWebRTCTransport extends Transport {
...
+          `webrtcUrl is deprecated, use webrtcRequestParams instead: { webrtcRequestParams: { endpoint: ${opts.webrtcUrl} } }`,
+        );
+      }
+      if (opts.webrtcRequestParams) {
+        logger.warn(
+          "Both webrtcUrl/connectionUrl and webrtcRequestParams provided. Using webrtcRequestParams.",
+        );
+      } else {
+        if (typeof _webrtcUrl === "string") {
+          this._webrtcRequest = { endpoint: _webrtcUrl };
```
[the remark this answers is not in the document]
mattieruth: nope. these will be filled in at the time of the request. filling in anything else would duplicate code or potentially override defaults that are better defined elsewhere. Also, where are you even getting your information from? the code you suggest does not match the type expected here.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -473,20 +541,27 @@ export class SmallWebRTCTransport extends Transport {
...
+      } = {
+        sdp: offerSdp.sdp,
+        type: offerSdp.type as string,
+        pc_id: this.pc_id,
+        restart_pc: recreatePeerConnection,
+      };
+      if (this._webrtcRequest.requestData) {
+        requestData.requestData = this._webrtcRequest.requestData;
+      }
+      request.requestData = requestData;
```
mattieruth, line 560: this logic here means that client-provided `requestData` will be nested under the field `requestData` such that the server endpoint will look like:

```
@app.post("/api/offer")
async def offer(request: dict, background_tasks: BackgroundTasks):
    client_data = request.get("requestData")
    if client_data:
        doCustomThing(client_data)
    pc_id = request.get("pc_id")
    ...
 ```
 
 So the question is: Do we like the naming of this field? should it just be `data`? or maybe `customData` or clientData`?

Comment (filipi87) [reply transports/small-webrtc-transport/src/smallWebRTCTransport.ts:560]: Names are always hard. But one thing I like about keeping it as `requestData` is that it matches what we are sending from the client side.

File: transports/small-webrtc-transport/package.json [hunk 4]
```diff
@@ -22,18 +22,20 @@
...
+    "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js",
+    "@types/lodash": "^4.17.20",
     "@types/node": "^22.9.0",
     "eslint": "9.11.1",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-simple-import-sort": "^12.1.1"
   },
   "peerDependencies": {
-    "@pipecat-ai/client-js": "~1.2.0"
+    "@pipecat-ai/client-js": "file:../../../pipecat-client-web/client-js"
```
filipi87, line 33: Same here.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -114,12 +122,40 @@ export class SmallWebRTCTransport extends Transport {
...
+        );
+      } else {
+        if (typeof _webrtcUrl === "string") {
+          this._webrtcRequest = { endpoint: _webrtcUrl };
+        } else {
+          logger.warn(`webrtcUrl is not a string. Ignoring: ${_webrtcUrl}`);
+        }
+      }
+    }
+    if (opts.webrtcRequestParams) {
```
filipi87, line 151: The only case where this wouldn’t be true is if we haven’t received either webrtcRequestParams, webrtcUrl, or connectionUrl, right ? 

But I guess that’s okay, since we can receive it directly in the connect method, right ?

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -221,12 +267,29 @@ export class SmallWebRTCTransport extends Transport {
 
     this.state = "connecting";
 
-    this._webrtcUrl =
-      connectParams?.webrtcUrl ??
-      connectParams?.connectionUrl ??
-      this._webrtcUrl;
-    if (!this._webrtcUrl) {
-      logger.error("No url provided for connection");
+    const _webrtcUrl =
```
filipi87, line 270: This logic seems to duplicate what we already do in the constructor. Could we unify it ? (checking webrtcUrl/connectionUrl, printing warnings, creating _webrtcRequest)

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
filipi87: Now, looking at the code above, it does feel weird to have `requestData.requestData`. It almost confused me until I realized that it’s actually the custom data from the user that we’re sending inside the request.

So, I am open to be convinced to use a different name. 😅

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
mattieruth: nm. i'm doing it.

matrunchyk:
> Note: This is ready for review, but note: it can not be merged until [pipecat-ai/pipecat-client-web#144](https://github.com/pipecat-ai/pipecat-client-web/pull/144) is merged and released.

Seems like it was released. Looking forward to merging this PR too.
And thank you for this PR! <3

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -221,12 +263,13 @@ export class SmallWebRTCTransport extends Transport {
...
-      connectParams?.webrtcUrl ??
-      connectParams?.connectionUrl ??
-      this._webrtcUrl;
-    if (!this._webrtcUrl) {
-      logger.error("No url provided for connection");
+    // Note: There is no need to validate the params here, as they were already
+    //       validated and fixed in the parent class's connect() method (which calls
+    //       _validateConnectionParams() and passes the result to _connect()).
+    this._webrtcRequest =
+      connectParams?.webrtcRequestParams ?? this._webrtcRequest;
```
filipi87, line 270: Just confirming: we don't need to invoke `_resolveRequestInfo` again here because we already did it inside `_validateConnectionParams`, right?

filipi87 (approved):
LGTM!.
Nice improvement. 🚀

PR #66, opened 2025-10-03
Improving how we wait for ICE gathering to complete.

filipi87:
Improving how we are handling for ice gathering to complete.


markbackman (approved):
This seems logical. It also works for me.

PR #71, opened 2025-10-16
Allowing startBotAndConnect to work with Pipecat cloud and Pipecat runner.

filipi87:
Allowing startBotAndConnect to work with Pipecat cloud and Pipecat runner.

### DIFF
--- a/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
+++ b/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
@@ -221,40 +221,115 @@ export class SmallWebRTCTransport extends Transport {
     return requestInfo ?? this._webrtcRequest;
   }
 
-  _validateConnectionParams(
-    connectParams: unknown,
-  ): SmallWebRTCTransportConnectionOptions | undefined {
-    if (connectParams === undefined || connectParams === null) {
-      return undefined;
+  _getStartEndpointAsString(): string | undefined {
+    const startEndpoint = this.startBotParams?.endpoint;
+    switch (typeof startEndpoint) {
+      case "string":
+        return startEndpoint;
+      case "object":
+        if (startEndpoint instanceof URL) {
+          return startEndpoint.toString();
+        }
+        if (startEndpoint instanceof Request) {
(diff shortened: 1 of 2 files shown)


markbackman:
I'm less familiar with this code, so I'll defer to @mattieruth.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -221,40 +222,125 @@ export class SmallWebRTCTransport extends Transport {
...
+      "webrtcUrl",
+      "connectionUrl",
+      "webrtcRequestParams",
+      "iceServers",
+    ];
+
+    if (this._shouldUseStartBotFallback(params, supportedKeys)) {
+      return this._buildConnectionOptionsBasedOnStartBotParams(params);
+    }
+
```
mattieruth: hmm. support for `iceServers` is now only actually supported in the version of this where you call `startBot()` and our runners. Following on my earlier comments, i think what's getting to me is that I feel like this function should look more like:

```
    if (!this._isValidObject(connectParams)) return undefined;

    const params = connectParams as Record<string, any>;

    const supportedKeys = [
      "webrtcUrl",
      "connectionUrl",
      "webrtcRequestParams",
      "iceServers",
    ];
    
    // where _fixConnectionOptionsParams does snake_cake -> camelCase changes
    // AND pulls the iceServers field out of the iceConfig, thus essentially just reformatting the params
    const fixedParams = this._fixConnectionOptionsParams(params, supportedKeys);

    // where the if(this._shouldUseStartBot() { this.buildConnectionOptions() }) is moved into 
    // `_resolveRequestInfo()` and `buildConnetionoptions` just returns the `webrtcRequestParams`
    // making it such that we have the one function for determining the proper value of `webrtcRequestParams`
    const webrtcRequestParams = this._resolveRequestInfo(fixedParams);
    ...
```

mattieruth (approved):

filipi87:
Closing this PR in favor of this one:
- https://github.com/pipecat-ai/pipecat-client-web-transports/pull/72

PR #73, opened 2025-11-17
Fixed an issue in `startBotAndConnect` where `requestData` was not carried over when invoking the `connect` endpoint.

filipi87:
Fixed an issue in `startBotAndConnect` where `requestData` was not carried over when invoking the `connect` endpoint.


jptaylor (approved):
LGTM

PR #78, opened 2025-12-01
Bumped min client-js version to 1.5.0 in all transports and bumped tr…

mattieruth:
…ansport versions in preparation for release


markbackman (approved):
LGTM

PR #82, opened 2026-01-12
Fix a bug in Firefox preventing microphone audio from being recorded …

kompfner:
…properly.

Tested with Firefox, Chrome, Brave, and Safari.

### DIFF
--- a/lib/wavtools/lib/mediastream_recorder.js
+++ b/lib/wavtools/lib/mediastream_recorder.js
@@ -3,6 +3,39 @@ import { AudioAnalysis } from "./analysis/audio_analysis.js";
 import { WavPacker } from "./wav_packer.js";
 
 /**
+ * Resamples audio data from one sample rate to another using linear interpolation
+ * @param {ArrayBuffer} inputBuffer - Input audio data as Int16 PCM
+ * @param {number} inputSampleRate - Sample rate of input audio
+ * @param {number} outputSampleRate - Desired output sample rate
+ * @returns {ArrayBuffer} - Resampled audio data as Int16 PCM
+ */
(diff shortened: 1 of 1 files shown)


lib/wavtools/lib/mediastream_recorder.js
```diff
@@ -135,7 +168,11 @@ export class MediaStreamRecorder {
 
     this.stream = new MediaStream([audioTrack]);
 
-    const context = new AudioContext({ sampleRate: this.sampleRate });
+    // Create AudioContext without specifying sample rate to avoid Firefox error:
+    // "Connecting AudioNodes from AudioContexts with different sample-rate is currently not supported."
+    // We'll resample audio from native rate to this.sampleRate when processing chunks.
```
kompfner: We could alternatively branch the code here, with one branch doing things "the Firefox way"—manually resampling—and the other branch doing things "the other-browsers way"—specifying `sampleRate` in the `AudioContext` constructor.

But I figured it'd be best from a maintainability standpoint to avoid bifurcating this logic, and revisiting this if we eventually decide it's worth the performance optimization of using the constructor parameter where possible.

WDYT?

mattieruth: sorry for the offensively long run-on sentence.

kompfner: OK, updated the logic to bifurcate between Firefox and other browsers.

mattieruth (approved):
looks good. trusting you that the rates match on non-firefox browsers, though, that makes sense.

filipi87 (approved):
LGTM! 🚀

PR #87, opened 2026-01-22
Fix issues surrounding requestData in the offer endpoint

mattieruth:
1. If the endpoint was a Request type, any requestData provided would be lost
2. Now that runners cache the requestData provided to /start, we do not need to re-send it in the offer (which conveniently solves issues if the endpoint provided to start was a Request type :))

This PR pairs nicely with https://github.com/pipecat-ai/pipecat-client-web/pull/171 -- though, not _technically_ dependent on it.


filipi87 (approved):
Thank you for catching this issue. 🚀🙌

PR #90, opened 2026-01-30
Aligning prettier config across client repos

mattieruth:
- Set trailingComma to "es5"
- Ran prettier across all files to ensure formatting is aligned


filipi87 (approved):
LGTM! 🚀

PR #113, opened 2026-04-24
test: characterize initDevices() across all transports

Regaddi:
## Summary

Plan A, step 1 (part 2 of 2). Adds a shared vitest test workspace and per-transport characterization specs ahead of the `MediaState` refactor in step 2. Companion to [pipecat-client-web#202](https://github.com/pipecat-ai/pipecat-client-web/pull/202).

Context: [Pipecat: Fixing the Device Picker Spinner After Disconnect](https://www.notion.so/dailyco/Pipecat-Fixing-the-Device-Picker-Spinner-After-Disconnect-3488fc5824d0810cb7ced697f3956ee6).

### New `tests/` workspace

- Added to root `workspaces` array; root `npm run test` delegates to `vitest run` in this workspace.
- `vitest.config.ts` aliases each transport package name to its `src` entry so characterization rides source, not dist.
- Uses `happy-dom` for the DOM environment.

### Shared helpers (`tests/src/helpers`)

- `observeTransport.ts`: builds a full spy-backed `RTVIEventCallbacks`, exposes a `StateRecorder` that captures every `onTransportStateChanged` emission, and wires a transport like `PipecatClient` does.
- `fakeMediaManager.ts`: structural MediaManager double for the DI-friendly transports (SmallWebRTC, WebSocket, Gemini Live). Tracks `initializeCallCount` and supports `initializeShouldThrow`.
- `fakeDaily.ts`: daily-js call-object stub covering the slice used by DailyTransport + OpenAIRealTimeWebRTCTransport during `initialize()` and `initDevices()`.

### Per-transport specs (5 files, 25 tests)

Each transport gets:
- initial state is `"disconnected"`
- happy-path `initDevices()` transitions (`disconnected` → `initializing` → `initialized`)
- repeated calls re-enter `"initializing"` (NOT idempotent at the transport layer)
- rejection leaves state at `"initializing"`

Daily and OpenAI additionally cover:
- fan-out of `onAvailable*Updated` / `on*Updated` exactly once per call
- partial internal idempotency around the audio-level observers (they guard restart via `isLocalAudioLevelObserverRunning`) — step 2 should preserve this when it short-circuits repeat calls at the client layer

No production code changes. No transport behavior is modified.

## Test plan

- [x] `npm run test` at repo root — all 25 tests pass across 5 spec files
- [x] `npx tsc --noEmit` in `tests/` — clean

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,27 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+
+      - name: Install dependencies
+        run: npm install --ignore-scripts
+
+      - name: Run tests
+        run: npm run test
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,8 @@
       "name": "pipecat-client-web-transports",
       "version": "0.0.0",
       "workspaces": [
-        "transports/*"
+        "transports/*",
+        "tests"
       ],
       "devDependencies": {
         "@commitlint/cli": "^20.0.0",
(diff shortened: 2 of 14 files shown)


mattieruth (approved):
this is some fun scaffolding. i think it looks right. these 'fake' classes always make me wary to try and make sure they actually properly mock the real thing, but i think they do. i have the same comment as i did on the client wrt maybe adding some more tests around toggling devices and initializing with the various devices disabled. Approving so that i'm not blocking

Regaddi:
The "initializing with the various devices disabled" half landed in 8cf9f6e (just after your approval). Each transport spec now has a parametrized `initialize(%j)` block covering all four `enableMic`/`enableCam` combinations:

- Daily and OpenAI: assert the options reach `startCamera` as `startAudioOff`/`startVideoOff`. OpenAI hardcodes `startVideoOff: true` regardless of `enableCam`, so we lock that quirk in with a comment pointing at Plan B3.
- SmallWebRTC, WebSocket, Gemini Live: assert the options reach `mediaManager.setClientOptions`.

The "toggling devices" half is deliberately deferred. Toggling `enableMic`/`enableCam` before connect or after `initDevices()` is the territory of Plan B1 (stale enable state in client-react after disconnect) and Plan B2 (Daily silently no-ops `enableMic`/`enableCam` when not joined). I'd rather characterize those next to the fixes so the assertions and the production change land in the same diff. Tracking it as a separate Plan B prep PR.

On the fakes: agreed, that's the standing risk with this kind of harness. The mock surface is intentionally narrow (state transitions, callback fan-out, options propagation) and each spec asserts against the real concrete transport's source, so we should notice the moment a real call signature drifts from what the fake returns. Worth a re-audit when step 2 lands and we add MediaState assertions on top.

PR #116, opened 2026-05-07
fix(ci): use paths_released to gate publish jobs in release-please workflow

Regaddi:
## Problem

The release-please workflow gated each publish job on a per-component output like `daily-transport--release_created`. These outputs are not reliably emitted when release-please processes multiple components in a single run. This caused publish jobs to be skipped even when release-please had just created a release for that component.

## Fix

Added `paths_released` to the `release-please` job's `outputs:` block and updated every publish job's `if:` condition to use `contains(fromJson(needs.release-please.outputs.paths_released), '<path>')`. The `paths_released` output is a JSON array of every package path released in the run, so it is always populated when a release occurs.

See: https://github.com/googleapis/release-please-action#outputs

## Jobs updated

- `publish-daily`: `transports/daily`
- `publish-gemini`: `transports/gemini-live-websocket-transport`
- `publish-openai`: `transports/openai-realtime-webrtc-transport`
- `publish-small-webrtc`: `transports/small-webrtc-transport`
- `publish-websocket`: `transports/websocket-transport`

The old per-component outputs are left in place (no removals) to avoid breaking anything that may reference them elsewhere.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -29,7 +29,7 @@ jobs:
 
   publish-daily:
     needs: release-please
-    if: ${{ contains(fromJson(needs.release-please.outputs.paths_released), 'transports/daily') }}
+    if: ${{ needs.release-please.outputs.releases_created == 'true' && contains(fromJson(needs.release-please.outputs.paths_released), 'transports/daily') }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -47,7 +47,7 @@ jobs:
 
   publish-gemini:
     needs: release-please
-    if: ${{ contains(fromJson(needs.release-please.outputs.paths_released), 'transports/gemini-live-websocket-transport') }}
+    if: ${{ needs.release-please.outputs.releases_created == 'true' && contains(fromJson(needs.release-please.outputs.paths_released), 'transports/gemini-live-websocket-transport') }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       releases_created: ${{ steps.release.outputs.releases_created }}
+      paths_released: ${{ steps.release.outputs.paths_released }}
```
filipi87, line 17: Is `paths_released` an supported output in `googleapis/release-please-action@v4` of the action ?

Regaddi: According to https://github.com/googleapis/release-please-action/tree/v4#outputs, yes!

filipi87 (approved):
LGTM!

.github/workflows/release-please.yml
```diff
@@ -28,7 +29,7 @@ jobs:
 
   publish-daily:
     needs: release-please
-    if: ${{ needs.release-please.outputs.daily_release_created == 'true' }}
+    if: ${{ contains(fromJson(needs.release-please.outputs.paths_released), 'transports/daily') }}
     runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
Regaddi: Fixed in a4c26d3. All 5 publish jobs now prepend `needs.release-please.outputs.releases_created == 'true' &&` before the `fromJson` call, so the condition short-circuits on non-release runs and `fromJson` is never evaluated against an empty/undefined value.

PR #129, opened 2026-05-21
Adding support for `DailyTransport` to receive audio faster than real time.

filipi87:
## Summary
  
- Added faster-than-realtime audio support for the Daily transport via a new `BotAudioPlayer` class — Pipecat sends true-rate PCM (e.g. 24 kHz) declared  as a higher rate (e.g. 48 kHz), so audio arrives ~2x faster than real-time; `BotAudioPlayer` captures that track at the declared rate and replays it through a `WavStreamPlayer` AudioContext running at the true rate, producing a correctly-pitched output track
- Added per-second buffer logging (`[BotAudioPlayer] Buffer: ~N samples (X.XXXs)`) and interrupt drop logging to aid debugging
- Improvements to make it work:
  - Fixed Chrome WebRTC audio pipeline not activating for Web Audio capture:
    Chrome requires a remote WebRTC track to be rendered in an `<audio>` element
    before `createMediaStreamSource()` receives audio; a hidden, volume-0 element
    now activates the pipeline without audible output
  - Fixed silence accumulation bug: WebRTC injects silence between utterances at
    the declared rate but the drain runs at the true rate, causing the buffer to
    grow unboundedly; silence frames are now discarded when the buffer is above
    a 100 ms threshold (matching the Python-side `_buffer_audio` logic), and
    kept below the threshold for pre-buffer smoothness

## Usage
  
Enable on the client side by passing `fasterThanRealtime: true` to the `DailyTransport` options: 
  
  ```ts
  const transport = new DailyTransport({
    fasterThanRealtime: true,
  });
  
On the Pipecat server side, configure the pipeline to use a declared sample rate higher than the true content rate (e.g. audio_out_declared_sample_rate=48000 with audio_out_sample_rate=24000) so audio is delivered faster than real-time over WebRTC.
  ```

## Related PRs
- Pipecat [PR](https://github.com/pipecat-ai/pipecat/pull/4544).
- New example [PR](https://github.com/pipecat-ai/pipecat-examples/pull/213).

### DIFF
--- a/transports/daily/src/bot_audio_player.ts
+++ b/transports/daily/src/bot_audio_player.ts
@@ -1,0 +1,82 @@
+import { WavStreamPlayer } from "../../../lib/wavtools";
+import { CaptureProcessorSrc } from "../../../lib/wavtools";
+
+/**
+ * Plays bot audio delivered faster-than-realtime over WebRTC.
+ *
+ * Pipecat writes true-rate PCM (e.g. 24 kHz) into a CustomAudioSource
+ * declared at a higher rate (e.g. 48 kHz), so audio arrives ~2x faster than
+ * real-time. This class captures that track at the declared rate, converts
+ * each render quantum to Int16 via an AudioWorklet, and feeds the samples
+ * into WavStreamPlayer whose AudioContext runs at the true rate — producing
+ * a correctly-pitched MediaStreamTrack that can be played normally.
+ */
+export class BotAudioPlayer {
+  private _captureCtx: AudioContext | null = null;
+  private _captureNode: AudioWorkletNode | null = null;
+  private _wavPlayer: InstanceType<typeof WavStreamPlayer> | null = null;
+
+  /**
+   * Start capturing and playing the bot's audio track.
+   *
+   * @param track           The bot's MediaStreamTrack from Daily (declared at declaredRate).
+   * @param declaredRate    The sample rate Pipecat declared to the CustomAudioSource (e.g. 48000).
+   * @param trueRate        The actual content sample rate of the audio (e.g. 24000).
+   * @returns               A MediaStreamTrack at trueRate with correctly-pitched audio.
+   */
+  async start(
+    track: MediaStreamTrack,
+    declaredRate: number,
+    trueRate: number
+  ): Promise<MediaStreamTrack> {
+    // Playback context at true rate. outputToSpeakers:false so audio only
+    // flows to outputTrack — the caller's <audio> element plays it, avoiding
+    // double playback from both context.destination and the audio element.
+    this._wavPlayer = new WavStreamPlayer({ sampleRate: trueRate, outputToSpeakers: false });
+    await this._wavPlayer.connect();
+
+    // Capture context at declared rate. sinkId:{type:'none'} prevents any
+    // audio from reaching the speakers from this context.
+    this._captureCtx = new AudioContext({
+      sampleRate: declaredRate,
+      // @ts-expect-error -- sinkId is not in all TypeScript lib versions yet
+      sinkId: { type: "none" },
+    });
+    await this._captureCtx.audioWorklet.addModule(CaptureProcessorSrc);
+
+    const source = this._captureCtx.createMediaStreamSource(
+      new MediaStream([track])
+    );
+    this._captureNode = new AudioWorkletNode(
+      this._captureCtx,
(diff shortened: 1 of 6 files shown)


transports/daily/src/transport.ts
```diff
@@ -47,11 +48,18 @@ export interface DailyConnectionEndpoint {
...
+   * through a WavStreamPlayer running at the true sample rate. This is used
+   * together with Pipecat's audio_out_declared_sample_rate feature, which
+   * delivers audio faster-than-realtime over WebRTC.
+   */
+  fasterThanRealtime?: boolean;
 }
 
 export enum DailyRTVIMessageType {
-  AUDIO_BUFFERING_STARTED = "audio-buffering-started",
-  AUDIO_BUFFERING_STOPPED = "audio-buffering-stopped",
```
mattieruth, line 54: were these not used before? -- either way, these would ideally be introduced to the client-web package. We don't otherwise have our RTVI definitions spread and we would need to bump the RTVI version

transports/daily/src/transport.ts
```diff
@@ -771,6 +827,34 @@ export class DailyTransport extends Transport {
       if (ev.participant?.local && ev.track.kind === "audio") {
         void this.handleLocalAudioTrack(ev.track);
       }
+
+      if (
+        this._fasterThanRealtime &&
+        !ev.participant?.local &&
+        ev.track.kind === "audio"
```
mattieruth, line 834: should this be kind === "audio" or type === "audio"? -- or rather, related to my comment on the pipecat PR, it comes down to whether this setting should be for all audio or mic-only (or specified per track).

transports/daily/src/transport.ts
```diff
@@ -630,6 +658,34 @@ export class DailyTransport extends Transport {
...
+          this._declaredSampleRate = ev.data.data.declared_sample_rate as number;
+          // Bot track arrived before rates were known — set up now and fire the
+          // deferred onTrackStarted with the correctly-pitched outputTrack.
+          if (this._pendingBotTrack) {
+            const track = this._pendingBotTrack;
+            const participant = this._pendingBotParticipant;
+            this._pendingBotTrack = null;
+            this._pendingBotParticipant = null;
+            void this._setupBotAudioPlayer(track).then(() => {
+              if (this._botOutputTrack) {
```
mattieruth, line 673: is there a scenario where a track is not returned? if so, shouldn't we handle that in some way? or be explicit about not handling it?

transports/daily/src/bot_audio_player.ts
```diff
@@ -0,0 +1,168 @@
...
+        this._maybeLogBufferStatus();
+      }
+    };
+
+    source.connect(this._captureNode);
+
+    return this._wavPlayer.outputTrack as MediaStreamTrack;
+  }
+
+  /** Clear the playback queue, mimicking the avatar stopping mid-speech. */
```
mattieruth, line 119: i'm guessing we have to handle this because the audio is coming in so fast that it's like the transcription arriving before the interrupt? is that true?

also, how do we un-interrupt? or is interruption simply that we strip whatever bytes we haven't read yet and continue-on, no need to "un-interrupt"

mattieruth (commented):
at a high level, i sort of asked this in the pipecat review, but... i think i'm missing the use case. what purpose is all this serving?

also, I essentially put this in my comments, but I think this requires a client-js change to add the new RTVI message type to the official types and bump the version.

filipi87:
Hi @mattieruth, this was more of a proof of concept. The idea originated from a requirement in Pipecat → `daily-python`, where the `daily-python` client ideally needed to receive audio faster than real time in order to buffer it and pre process video avatar images, so that both the image and audio would be ready to send back to the user at the playback rate.

This would give us the advantage of being able to pre process the audio before using it downstream.

Then, after discussing it with Kwin, we thought it would be interesting to test this in the browser to see whether it could help compensate for temporary poor network conditions.

In the end, though, there are a couple of limitations, which I detailed in the example:
- https://github.com/pipecat-ai/pipecat-examples/tree/filipi/daily_faster_than_realtime.py/faster-than-realtime#known-limitations

That is also why I haven’t put much effort into designing a proper RTVI protocol for it. I left it as a draft so we would know that there is a viable way to make it work, along with what it would require.

I also explained all the findings in detail in a Slack thread, I will share with you. 🙂

PR #137, opened 2026-06-01
Log a warning instead of an error for unsupported features

filipi87:
Log a warning instead of an error for unsupported features


markbackman (approved):
👍

mattieruth (approved):

PR #146, opened 2026-06-08
feat: support optional auth token in connection params

filipi87:
feat: support optional auth token in connection params

### DIFF
--- a/transports/websocket-transport/src/webSocketTransport.ts
+++ b/transports/websocket-transport/src/webSocketTransport.ts
@@ -126,7 +126,8 @@ export class WebSocketTransport extends Transport {
 
     this._wsUrl = connectParams?.wsUrl ?? connectParams?.ws_url ?? this._wsUrl;
     if (connectParams?.token) {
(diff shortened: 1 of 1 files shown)


transports/websocket-transport/src/webSocketTransport.ts
```diff
@@ -118,6 +125,9 @@ export class WebSocketTransport extends Transport {
     this.state = "connecting";
 
     this._wsUrl = connectParams?.wsUrl ?? connectParams?.ws_url ?? this._wsUrl;
+    if (connectParams?.token) {
+      this._wsUrl = `${this._wsUrl}?token=${encodeURIComponent(connectParams.token)}`;
```
mattieruth: is there any chance wsUrl already has a `?`. should we have logic to detect and use a `&` instead?

filipi87: I think it doesn't hurt to be extra careful, so I'll add this logic. 👍

filipi87: Done.

mattieruth (approved):

PR #139, opened 2026-06-02
add moq transport support

vipyne:
Part of MoQ transport addition: https://github.com/pipecat-ai/pipecat/pull/4629

PR #148, opened 2026-06-17
MoQ: play bot audio via @moq/watch buffered playback

kixelated:
Stacked on top of #139 (targets `vp-add-moq-transport`).

## What

Replaces the hand-rolled bot-audio playback (`@moq/hang` `Container.Consumer` + a manual back-to-back WebAudio scheduler) with `@moq/watch`'s audio chain:

`Watch.Broadcast → Audio.Source → Audio.Decoder → Audio.Emitter`, driven by a `Sync` configured for buffered playback.

## Why

The bot now writes TTS faster than real-time with future-dated PTS (see the paired pipecat change). The old `_playAudioData` scheduler appended decoded chunks on the AudioContext clock — it buffered, but couldn't handle loss and ran at higher latency. `@moq/watch` decodes into an AudioWorklet ring buffer with timestamp-aware backpressure, plays at the encoded pace, and supports a clean flush.

## Changes
- `_consumeBotAudio`/`_playAudioData` → the `@moq/watch` chain; the decoder's worklet is tapped into a `MediaStreamDestination` for `tracks().bot.audio`.
- On interruption (`user-started-speaking` from the transcript track), `reset()` flushes the buffer so the previous utterance stops immediately.
- New `audioBufferMaxMs` option (the `Sync` latency ceiling), default **5 min**.
- Adds the `@moq/watch` dependency.

Depends on the watch fix in [moq #1620](https://github.com/moq-dev/moq/pull/1620) (consumer sized by the latency ceiling). `@moq/watch` is not yet on npm, so the lockfile entry is intentionally omitted here.

### DIFF
--- a/0001-what-will-merge.patch
+++ b/0001-what-will-merge.patch
@@ -1,0 +1,13 @@
+diff --git a/package.json b/package.json
+index 046412c..0a97d1b 100644
+--- a/package.json
++++ b/package.json
+@@ -28,6 +28,7 @@
+     "typescript": "^5.5.4"
+   },
+   "peerDependencies": {
+-    "@daily-co/daily-js": "^0.90.0"
++    "@daily-co/daily-js": "^0.90.0",
++    "@pipecat-ai/client-js": "^1.10.0"
+   }
+ }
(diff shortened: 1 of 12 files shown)


kixelated:
Superseded by #155, which lands the standalone JSON-stream transcript change (lossless `Json.Stream.Producer`/`Consumer` + compression, `transcript.json.z`) on a clean `vp-add-moq-transport` base, without the @moq/watch audio-playback WIP this branch carried. Closing in favor of #155.

PR #155, opened 2026-07-13
feat(moq): carry the RTVI transcript on a lossless compressed JSON stream

kixelated:
Stacked on #139 (base `vp-add-moq-transport`). Browser companion to pipecat **pipecat-ai/pipecat#5015**.

## What

Move both MoQ transcript directions from `@moq/json`'s **snapshot** mode
(`Json.Producer`/`Json.Consumer`, RFC 7396 merge-patch) to the **Stream**
mode (`Json.Stream.Producer`/`Json.Stream.Consumer`) — an ordered,
lossless append-log, with **compression on**. Track renamed
`transcript` → `transcript.json.z`.

## Why

Snapshot mode collapses to the latest value: a consumer that has fallen
behind (or joined late) catches up in one step and **silently drops the
intermediate RTVI events**. RTVI is a stream of discrete events
(`bot-output`, `client-ready`, function-call results, …), not a mutating
document, so those drops are lost messages. Stream mode delivers every
appended record, in order.

This is wire-compatible with, and the counterpart to, the bot's Rust
`moq_json::stream` on the Python side (#5015): the bot publishes with
`publish_json_stream(compression=True)` and subscribes to the client's
transcript with `subscribe_json_stream(compression=True)`.

## How

- **Consumer** (bot → client): `new Json.Stream.Consumer(track, { compression: true })`.
- **Producer** (client → bot): the stream `Producer` binds to a single
  track with no built-in fan-out, and `publishTrack` serves each
  subscriber its own track — so we hold one `Producer` per subscription
  and replay a message log into each. A bot that subscribes *after*
  `sendMessage` has already run still gets every message, in order.
- Bump `@moq/json` `^0.1.1` → `^0.1.2` (0.1.2 is the first release
  exposing the `Stream` namespace); re-lock. (The lock also picks up
  transitive updates that `vp-add-moq-transport`'s lock was already behind
  on within its existing ranges.)

## Testing

- `tsc --noEmit`: clean.
- Validated a lossless, compressed round-trip through
  `Json.Stream.Producer` → `Json.Stream.Consumer` (5 RTVI-shaped records
  appended, all 5 back in order).

Note: supersedes #148 (that branch carried this on top of unrelated
@moq/watch audio-playback WIP); this is the standalone JSON-stream change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -383,7 +383,6 @@
       "resolved": "https://registry.npmjs.org/@daily-co/daily-js/-/daily-js-0.90.0.tgz",
       "integrity": "sha512-XnuuNIxLt8GOv+rFYoU+OPTSmVj+Mg9hRPK2EM6WYVY62F6rcw7vwzZBOSCisVuu1VnTIF/2J9V0VEBa83lDzw==",
       "license": "BSD-2-Clause",
-      "peer": true,
       "dependencies": {
         "@babel/runtime": "^7.12.5",
         "@sentry/browser": "^8.33.1",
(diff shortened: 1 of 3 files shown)


vipyne (approved):
sounds good to me

PR #164, opened 2026-07-24
fix: Fix issue where the DailyMediaManager would not resolve in conne…

### LINKED ISSUE #154: SmallWebRTCTransport: second connect() on a reused client hangs before sending the offer
Using `@pipecat-ai/small-webrtc-transport` 1.10.5 with `client-js` 1.12.0, default transport setup (mic on, cam off), one `PipecatClient` reused across calls:

```js
const client = new PipecatClient({
  transport: new SmallWebRTCTransport(),
  enableMic: true,
  enableCam: false,
});
await client.connect({ webrtcRequestParams: { endpoint: OFFER_URL } }); // ok
await client.disconnect();
await client.connect({ webrtcRequestParams: { endpoint: OFFER_URL } }); // hangs forever
```

The second `connect()` never sends the offer (no getUserMedia, no fetch) and never resolves or rejects; the transport sits in `connecting`. Reproduces every time, in Chrome, Firefox and Safari.

What I found digging through the dist:

- The default media manager is constructed as `DailyMediaManager(false, false, ...)`, so `_mediaStreamRecorder` is undefined.
- `disconnect()` resets the manager's `_initialized`, but the client's mediaState stays `"granted"`, so `needsInit()` is false on reconnect and `initDevices()` is skipped.
- `DailyMediaManager.connect()` then takes the `!_initialized` path and waits on `_connectResolve`, whose only call site is in `handleTrackStarted` behind `if (!this._mediaStreamRecorder) return`. With recording disabled (the default) that resolver is unreachable, so the promise never settles and `_connect()` never reaches `startNewPeerConnection()`.

Related: `SmallWebRTCTransport.stop()` early-returns when `this.pc` is null, which is exactly the state the hung connect leaves behind. The following `disconnect()` therefore skips `mediaManager.disconnect()` and strands `_connected = true`, so the *next* `connect()` hits the "DailyMediaManager already connected" guard and accidentally works. That's why disconnecting and connecting a second time appears to fix it.

Also worth noting: the connect promise's executor fires `initialize()` without wiring its rejection anywhere, so a `startCamera()` failure on this path also hangs `connect()` instead of rejecting.

The workaround I'm using: `await client.initDevices()` before every `connect()`, which keeps the manager on the initialized fast path. It has survived repeated connect/disconnect cycles in my testing.

A fix could be to resolve the connect promise when `initialize()` completes (and reject on its failure) instead of relying on the recorder's track-started branch, and to have `stop()` still call `mediaManager.disconnect()` when `pc` is null.

mattieruth:
…ct() on second connect(). Fixes issue 154

### DIFF
--- a/lib/media-mgmt/dailyMediaManager.ts
+++ b/lib/media-mgmt/dailyMediaManager.ts
@@ -145,6 +145,12 @@ export class DailyMediaManager extends MediaManager {
         (async () => {
           this._connectResolve = resolve;
           await this.initialize();
+          // Without a recorder, handleTrackStarted never fires to resolve this
+          // promise, so resolve here once initialization is complete.
+          if (this._connectResolve && !this._mediaStreamRecorder) {
+            this._connectResolve();
+            this._connectResolve = null;
+          }
         })();
       });
     }


Regaddi (approved):

filipi87 (approved):
LGTM!

PR #167, opened 2026-08-03
MoQ: wait for the bot's broadcast before subscribing to it

kixelated:
## What

Both of the bot-side subscribers — the `Watch.Broadcast` catalog/audio watcher and the transcript stream — fired on `this._reload.established`, i.e. as soon as the *connection* came up. That assumes the bot is already publishing on the relay.

It isn't always. When the bot is started *in response to* the client's own announcement, its broadcast necessarily appears after ours. The relay resets a subscribe to a path nobody publishes, so both streams died with `RESET_STREAM` and neither retried, leaving the client connected but permanently silent:

```
subscribe start: id=0 broadcast=response/<id> track=transcript.json.z
subscribe start: id=1 broadcast=response/<id> track=catalog.json
announce:        broadcast=request/<id> active=true
subscribe error: id=1 ... track=catalog.json error=Received RESET_STREAM.
subscribe error: id=0 ... track=transcript.json.z error=Received RESET_STREAM.
announced:       broadcast=response/<id> active=true      <- bot arrives, too late
```

This doesn't affect the `/start` flow, where the server blocks on the bot's ready event before the browser ever connects.

## How

Track the bot's broadcast announcement in a `Signal` and gate both subscribers on it. `Watch.Broadcast` already accepts an `enabled` signal, and the transcript effect re-runs when the signal flips — so each subscribes once the broadcast is really there, and re-subscribes if it goes away and comes back.

## Testing

Verified against a bot that starts on demand, so its broadcast is guaranteed to appear after the client's. Before: both subscribes reset, no transcript, no audio. After: catalog fetch succeeds, transcript renders, `sync[audio]` shows the player consuming frames. `tsc --noEmit` and `parcel build` clean.

(written by Opus 5)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5857,7 +5857,6 @@
       "version": "7.8.2",
       "license": "Apache-2.0",
       "optional": true,
-      "peer": true,
       "dependencies": {
         "tslib": "^2.1.0"
       }
@@ -6569,7 +6568,7 @@
     },
     "transports/moq-transport": {
       "name": "@pipecat-ai/moq-transport",
-      "version": "0.0.1",
+      "version": "0.1.0",
       "license": "BSD-2-Clause",
       "dependencies": {
         "@moq/hang": "^0.2.12",
--- a/transports/moq-transport/src/moqTransport.ts
+++ b/transports/moq-transport/src/moqTransport.ts
@@ -578,12 +578,22 @@ export class MoqTransport extends Transport {
       );
     });
 
-    // Transcript — a lossless JSON append-stream over a single track. We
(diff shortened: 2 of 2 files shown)


vipyne (approved):

kixelated:
Oh and I'm working on a better API for this, because yeah it's kind of gross. But I'll submit a separate PR for that.

vipyne:
I pushed a commit, lmk if it works for you @kixelated. it's intended to be paired with [pipecat PR #5158](https://github.com/pipecat-ai/pipecat/pull/5158)

filipi87 (approved):
LGTM!

PR #86, opened 2026-01-20
feat: add livekit transport

monster-anshu:
Adds a new @pipecat-ai/livekit-transport package to the monorepo — a WebRTC transport for @pipecat-ai/client-js built on LiveKit's infrastructure. Unlike the peer-to-peer small-webrtc-transport, this routes media through LiveKit's SFU, enabling multi-participant rooms, edge routing, and production-scale deployments.

What's included

- LiveKitTransport (src/liveKitTransport.ts) — extends the base Transport class and implements the full Pipecat
transport lifecycle:
  - Connection — flexible auth: fetch { url, token } from an authUrl (GET or POST with body), or pass url/token
directly to connect(). Constructor accepts LiveKit RoomOptions.
  - Device management — enumerate/switch mics, cams, and speakers via navigator.mediaDevices + LiveKit's
switchActiveDevice; hot-swapping via a devicechange listener; selected-device state synced to callbacks.
  - Media — mic/cam enable/disable, screen share (with audio), local + remote track publish/subscribe wired to
onTrackStarted/onTrackStopped.
  - Messaging — RTVI messages over LiveKit data channels (reliable), encoded/decoded via TextEncoder/TextDecoder.
  - Participants — join/leave callbacks; first remote participant treated as the bot.
  - State — mapped to Pipecat states (disconnected → ready/error), with abort-signal handling on connect.
- Package scaffolding — package.json, tsconfig.json, LICENSE (BSD-2-Clause), README.md, CHANGELOG.md, src/index.ts.
- Dependencies — livekit-client ^2.21.0; peer dep @pipecat-ai/client-js ^1.13.0.
- Docs — root README.md updated to list the new transport; package-lock.json regenerated.

PR #171, opened 2026-08-17
Monster anshu/86 add livekit

mattieruth:
From original PR #86:

Adds a new @pipecat-ai/livekit-transport package to the monorepo — a WebRTC transport for @pipecat-ai/client-js built on LiveKit's infrastructure. Unlike the peer-to-peer small-webrtc-transport, this routes media through LiveKit's SFU, enabling multi-participant rooms, edge routing, and production-scale deployments.

What's included

- LiveKitTransport (src/liveKitTransport.ts) — extends the base Transport class and implements the full Pipecat
transport lifecycle:
  - Connection — flexible auth: fetch { url, token } from an authUrl (GET or POST with body), or pass url/token
directly to connect(). Constructor accepts LiveKit RoomOptions.
  - Device management — enumerate/switch mics, cams, and speakers via navigator.mediaDevices + LiveKit's
switchActiveDevice; hot-swapping via a devicechange listener; selected-device state synced to callbacks.
  - Media — mic/cam enable/disable, screen share (with audio), local + remote track publish/subscribe wired to
onTrackStarted/onTrackStopped.
  - Messaging — RTVI messages over LiveKit data channels (reliable), encoded/decoded via TextEncoder/TextDecoder.
  - Participants — join/leave callbacks; first remote participant treated as the bot.
  - State — mapped to Pipecat states (disconnected → ready/error), with abort-signal handling on connect.
- Package scaffolding — package.json, tsconfig.json, LICENSE (BSD-2-Clause), README.md, CHANGELOG.md, src/index.ts.
- Dependencies — livekit-client ^2.21.0; peer dep @pipecat-ai/client-js ^1.13.0.
- Docs — root README.md updated to list the new transport; package-lock.json regenerated.

This PR will be followed up with:
- Voice UI Kit: https://github.com/pipecat-ai/voice-ui-kit/pull/159
- Pipecat Prebuilt:  https://github.com/pipecat-ai/pipecat-prebuilt/pull/52

### DIFF
--- a/transports/livekit-transport/src/liveKitTransport.ts
+++ b/transports/livekit-transport/src/liveKitTransport.ts
@@ -1,0 +1,557 @@
+import {
+  DeviceError,
+  Participant,
+  PipecatClientOptions,
+  RTVIEventCallbacks,
+  RTVIMessage,
+  Tracks,
+  Transport,
+  TransportStartError,
+  TransportState,
+  logger,
+} from "@pipecat-ai/client-js";
+import {
+  LocalParticipant,
+  LocalTrackPublication,
+  RemoteParticipant,
+  RemoteTrack,
+  RemoteTrackPublication,
+  Room,
+  RoomConnectOptions,
+  RoomEvent,
+  RoomOptions,
+  Track,
+} from "livekit-client";
+
+type LiveKitConnectParams = {
+  authUrl: string;
+  authMethod: "GET" | "POST";
+  authBody: Record<string, unknown>;
+
+  url: string;
+  token: string;
+
+  roomConnectionOptions: RoomConnectOptions;
+};
+
+export type LiveKitTransportConstructorOptions = RoomOptions;
+
+export class LiveKitTransport extends Transport {
+  private _room: Room;
+
+  private _selectedMic: MediaDeviceInfo | Record<string, never> = {};
+  private _selectedCam: MediaDeviceInfo | Record<string, never> = {};
+  private _selectedSpeaker: MediaDeviceInfo | Record<string, never> = {};
+
+  private _micEnabled: boolean = false;
+  private _camEnabled: boolean = false;
+  private _listenersAttached: boolean = false;
+  private _deviceChangeHandler = () => this.updateAvailableDevices();
+
+  protected _state: TransportState = "disconnected";
+  protected _callbacks: RTVIEventCallbacks = {};
+
+  constructor(options: LiveKitTransportConstructorOptions = {}) {
+    super();
+    this._room = new Room(options);
+  }
+
+  public initialize(
+    options: PipecatClientOptions,
+    messageHandler: (ev: RTVIMessage) => void
+  ): void {
+    this._options = options;
+    this._callbacks = options.callbacks ?? {};
+    this._onMessage = messageHandler;
+    this._micEnabled = options.enableMic ?? false;
+    this._camEnabled = options.enableCam ?? false;
+
+    this.attachEventListeners();
+
+    this.state = "disconnected";
+    logger.debug("[LiveKit Transport] Initialized");
+  }
+
+  get state(): TransportState {
+    return this._state;
+  }
+
+  set state(state: TransportState) {
+    if (this._state === state) return;
+
+    this._state = state;
+    this._callbacks.onTransportStateChanged?.(state);
+  }
+
+  async initDevices(): Promise<void> {
+    this.state = "initializing";
+
+    // In LiveKit, we can pre-warm devices or just enumerate
+    // We'll enumerate first
+    await this.updateAvailableDevices();
+
+    // Enable devices based on options if needed, but usually this happens on connect or explicitly
+    // For now, we just mark as initialized
+    this.state = "initialized";
+  }
+
+  private async updateAvailableDevices() {
+    // LiveKit helper for listing devices
+    // Note: LiveKit's Room doesn't strictly expose enumerateDevices directly as a static helper in all versions,
+    // but we can use navigator.mediaDevices directly or Room.getLocalDevices helper if available.
+    // We will use navigator.mediaDevices standard API for enumeration as it is robust.
+
+    try {
+      const devices = await navigator.mediaDevices.enumerateDevices();
+      const cams = devices.filter((d) => d.kind === "videoinput");
+      const mics = devices.filter((d) => d.kind === "audioinput");
+      const speakers = devices.filter((d) => d.kind === "audiooutput");
+
+      this._callbacks.onAvailableCamsUpdated?.(cams);
+      this._callbacks.onAvailableMicsUpdated?.(mics);
+      this._callbacks.onAvailableSpeakersUpdated?.(speakers);
+
+      // We can't easily know "selected" device without querying the Room's LocalParticipant or active device
+      // But initially none is selected until we start tracks.
+    } catch (e) {
+      logger.error("Error enumerating devices", e);
+    }
+  }
+
+  _validateConnectionParams(
+    connectParams: unknown
+  ): LiveKitConnectParams | undefined {
+    if (!connectParams || typeof connectParams !== "object") return undefined;
+    return connectParams as LiveKitConnectParams;
+  }
+
+  async _connect(connectParams?: LiveKitConnectParams): Promise<void> {
+    const params = connectParams || ({} as Partial<LiveKitConnectParams>);
+    let { url, token } = params;
+
+    this.state = "connecting";
+    if (params.authUrl) {
+      try {
+        const options: RequestInit = {
+          method: params.authMethod ?? "GET",
+          headers: {
+            "Content-Type": "application/json",
+          },
+        };
+        if (options.method?.toUpperCase() == "POST") {
+          options.body = JSON.stringify(params.authBody ?? {});
+        }
+        const res = await fetch(params.authUrl, options);
+        const json = await res.json();
+        url = json.url;
+        token = json.token;
+      } catch (e) {
+        logger.error("Failed to fetch LiveKit credentials from authUrl", e);
+        this.state = "error";
+        throw new TransportStartError("Failed to fetch credentials");
+      }
+    }
+
+    if (!url || !token) {
+      logger.error(
+        "LiveKit connection requires 'url' and 'token' or 'authUrl'"
+      );
+      this.state = "error";
+      throw new TransportStartError("Missing url or token");
+    }
+
+    try {
+      await this._room.connect(url, token, params.roomConnectionOptions);
+      const enableMic = this._micEnabled;
+      const enableCam = this._camEnabled;
+      await this._room.localParticipant.setMicrophoneEnabled(enableMic);
+      if (enableMic) {
+        const trackPub = this._room.localParticipant.getTrackPublication(
+          Track.Source.Microphone
+        );
+        if (trackPub?.track?.mediaStreamTrack) {
+          const deviceId =
+            trackPub.track.mediaStreamTrack.getSettings().deviceId;
+          if (deviceId) {
+            const mics = await this.getAllMics();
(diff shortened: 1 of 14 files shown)


vipyne (approved):
tested locally and worked for me.  some thoughts here: https://github.com/pipecat-ai/pipecat-client-web-transports/pull/172 but don't want to block

transports/livekit-transport/src/liveKitTransport.ts
filipi87, line 1: Are we going to add it to the release please ?

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+      this._callbacks.onSpeakerUpdated?.(speaker);
+    }
+  }
+
+  private async _acquireMic(): Promise<void> {
+    try {
+      // See the comment in initDevices() re: { deviceId: "default" } — a
+      // soft ideal preference, not exact, so this stays safe on Firefox/Safari.
+      await this._adoptMicTrack(
+        await createLocalAudioTrack({ deviceId: "default" })
```
filipi87, line 172: If we have a `this._selectedMic` defined, shouldn't we use it instead of always using the `default` ?

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+          e instanceof Error ? e.message : String(e)
+        )
+      );
+    }
+  }
+
+  private async _acquireCam(): Promise<void> {
+    try {
+      await this._adoptCamTrack(
+        await createLocalVideoTrack({ deviceId: "default" })
```
filipi87, line 189: Same here for camera. Should we use the `this._selectedCam` ?

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+        const videoTrack = tracks.find((t) => t.kind === Track.Kind.Video) as
+          | LocalVideoTrack
+          | undefined;
+        if (audioTrack) await this._adoptMicTrack(audioTrack);
+        if (videoTrack) await this._adoptCamTrack(videoTrack);
+      } catch {
+        // createLocalTracks, like getUserMedia, is all-or-nothing — one
+        // device failing fails the whole combined request. Fall back to
+        // acquiring each independently so a single bad device doesn't take
+        // the other down with it, and the error attributes to the right one.
```
filipi87, line 130: Should we show a warning if we fall back to this case?

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+  }
+
+  private async updateAvailableDevices(): Promise<
+    MediaDeviceInfo[] | undefined
+  > {
+    try {
+      const devices = await navigator.mediaDevices.enumerateDevices();
+      const cams = devices.filter((d) => d.kind === "videoinput");
+      const mics = devices.filter((d) => d.kind === "audioinput");
+      const speakers = devices.filter((d) => d.kind === "audiooutput");
```
filipi87: We invoke devices in several places, would it be worth to centralize it in a single method ? Something like this:
```
private async _enumerateDevices(): Promise<{
  all: MediaDeviceInfo[];
  mics: MediaDeviceInfo[];
  cams: MediaDeviceInfo[];
  speakers: MediaDeviceInfo[];
}> {
  const all = await navigator.mediaDevices.enumerateDevices();
  return {
    all,
    mics: all.filter(d => d.kind === "audioinput"),
    cams: all.filter(d => d.kind === "videoinput"),
    speakers: all.filter(d => d.kind === "audiooutput"),
  };
}
```

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+   * On devicechange, check whether the mic/cam we're actively capturing from
+   * is still present (or, for a "default" selection, whether the underlying
+   * physical default changed). livekit-client's own Room.selectDefaultDevices
+   * already does this for audiooutput on every devicechange (and switching
+   * the active output there also updates every remote participant's <audio>
+   * element — see updateSpeaker()), but it deliberately skips audioinput
+   * (except a Safari AirPods special case) and videoinput, leaving mic/cam
+   * reselection to us.
+   */
+  private async _handleDeviceChange(): Promise<void> {
```
filipi87, line 244: I guess removing a headset can affect multiple device types at the same time, or not? Just wondering if we need to implement some kind of serialization here in case that happens.

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+  }
+
+  // updateMic/updateCam switch the device on our own owned LocalAudioTrack/
+  // LocalVideoTrack via restartTrack() — this works whether or not the track
+  // is currently published (Room-agnostic re-acquire pre-connect; a smooth
+  // sender.replaceTrack() with no renegotiation once published). There's
+  // nothing to switch if the device was never enabled in the first place.
+  async updateMic(micId: string): Promise<void> {
+    if ((this._selectedMic as MediaDeviceInfo).deviceId === micId) return;
+    if (!this._localAudioTrack) return;
```
filipi87: Should we show some warning in this case ? What we are basically saying here is, you can only invoke `updateMic`, if `initDevices` have been invoked, right ?

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+          this.toDeviceErrorType(MediaDeviceFailure.getFailure(e)),
+          e instanceof Error ? e.message : String(e)
+        )
+      );
+    }
+  }
+
+  async updateCam(camId: string): Promise<void> {
+    if ((this._selectedCam as MediaDeviceInfo).deviceId === camId) return;
+    if (!this._localVideoTrack) return;
```
filipi87: Same here.

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+  // depends on an active engine connection, createLocalAudioTrack() doesn't.
+  //
+  // When connected, always go through setMicrophoneEnabled() rather than
+  // branching on whether we already hold a track ourselves — that's exactly
+  // what it does internally (livekit-client's own setTrackEnabled() calls
+  // track.mute()/unmute() when a publication already exists, or creates +
+  // publishes a fresh one otherwise), so there's no reason to duplicate that
+  // decision on our end. Either way, _syncSelectedMic() afterwards figures
+  // out what actually changed and reports it.
+  async enableMic(enable: boolean): Promise<void> {
```
filipi87, line 624: This code is a little confusing. 

Could we try to simplify, like separating "perform the toggle" from "synchronize the resulting track." ?

I asked to Claude and it mentioned something like:

```
_setMicEnabled()
    ↓
"Make mic enabled/disabled"

_syncMicTrack()
    ↓
"Make our local track representation consistent"
```

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1061 @@
...
+  //   - device: compares against _selectedMic/_selectedCam, fires
+  //     onMicUpdated/onCamUpdated if the resolved device differs.
+  //   - validity: compares the current "live" raw track (undefined while
+  //     muted) against the last one we reported via onTrackStarted, firing
+  //     onTrackStopped/onTrackStarted for whatever actually differs. This
+  //     covers every case uniformly: a fresh acquire (nothing → track), a
+  //     mute (track → nothing), an unmute (nothing → track, same or restarted
+  //     object), a device switch (track → different track), and a no-op
+  //     repeat call (nothing to report either way).
+  private async _syncSelectedMic(track: LocalAudioTrack): Promise<void> {
```
filipi87: `_syncSelectedMic()` and `_syncSelectedCam() `are almost identical, would it be worth try to abstract ?

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: good catch. yes.

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: good idea.

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: _acquireMic/Cam are only called from `initDevices()` which occurs before the selectedMic is ever set. I guess we could default `_selectedMic` to {device: "default"}, but that might lose the ability to know if that value has been set to reality or not.

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: same response. i lean towards no.

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: are you concerned about one device event with multiple changes? or multiple device events occurring in quick succession? Seems like you're concerned about the former and if so, we should be safe. If the latter, claude points out you could technically end up calling restartTrack() on the same track concurrently. I don't really think that's a concern though.

mattieruth:
alright. in theory, this is now really ready! all feedback should be addressed and vanessa's changes are in.

transports/livekit-transport/src/liveKitTransport.ts
filipi87: Ok, I see. Agreed, in this case I think keeping the default is fine. 👍

transports/livekit-transport/src/liveKitTransport.ts
filipi87: Agreed. 👍

transports/livekit-transport/src/liveKitTransport.ts
filipi87: > device event with multiple changes

Exactly, this was the case I had in mind. I am just not sure if this can actually happen.

As for the multiple devices in succession case, I agree with you. It feels like a pretty edge case to worry about, at least for now.

transports/livekit-transport/src/liveKitTransport.ts
```diff
@@ -0,0 +1,1120 @@
...
+
+    // Re-attach the devicechange listener dropped by a prior _disconnect() —
+    // initialize() only runs once per client lifetime, so a reconnect on the
+    // same transport must restore it here. Re-adding an already-registered
+    // listener is a no-op per the DOM spec, so the first connect is
+    // unaffected.
+    navigator.mediaDevices.addEventListener(
+      "devicechange",
+      this._deviceChangeHandler
+    );
```
filipi87, line 343: Inside `_connect`, the first time we connect, we may end up registering this same event again, but I think it’s fine. I believe the browser will ignore the second registration.

filipi87 (approved):
LGTM! Thank you for bringing this all the way to the finish line. 🏁🚀

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: note: I renamed `_acquireMic/Cam` to `_initMic/Cam` to make this clearer.

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: good thought. i just confirmed that the browser does indeed override and only register one. so we're good.

transports/livekit-transport/src/liveKitTransport.ts
mattieruth: done.

PR #172, opened 2026-08-25
Fix bot identification, devicechange listener lifecycle, and ready ga…

vipyne:
…ting

Three review fixes, each with a characterization test added first:

1. Identify a bot already present in the room at join time. livekit-client only emits ParticipantConnected for participants who join after us: join-response participants are processed while the room is still "connecting", and Room.emitWhenConnected() silently drops (doesn't buffer) events in that state. With the pipecat runner the agent typically connects before the client, so _botId was never set -- onBotConnected/onBotDisconnected never fired and tracks().bot was always empty. _connect() now sweeps room.remoteParticipants after connecting, running each through the same handleParticipantConnected path the event-driven flow uses.

2. Re-attach the devicechange listener on reconnect. initialize() attaches it and _disconnect() removes it, but client-js only calls initialize() once per client lifetime -- so a disconnect -> connect cycle on the same transport silently lost device hot-plug handling. _connect() now re-adds the listener (re-adding an identical listener is a no-op per the DOM spec, so the first connect is unaffected).

3. Gate sendReadyMessage() on the bot's own media. handleTrackSubscribed resolved a pending ready promise on any track subscription from any participant; in a multi-participant room another human's track could mark the session ready before the bot published anything. The ready handler now only fires when the subscribed track belongs to _botId (which fix 1 guarantees is set before sendReadyMessage() runs).

### DIFF
--- a/tests/src/transports/livekit.spec.ts
+++ b/tests/src/transports/livekit.spec.ts
@@ -581,6 +581,42 @@ describe("LiveKitTransport — characterization", () => {
       ).toHaveBeenCalledWith(expect.any(Uint8Array), { reliable: true });
     });
 
+    test("sendReadyMessage() ignores track subscriptions from non-bot participants", async () => {
+      const { callbacks } = buildSpyCallbacks();
+      wireTransport(transport, callbacks);
+
+      const bot = {
+        identity: "bot-1",
+        name: "Bot",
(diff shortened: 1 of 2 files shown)


mattieruth:
nice! thanks! merging as-is (rebased on latest main to avoid issues)

PR #174, opened 2026-08-27
fix(small-webrtc): stop retrying refused offers and surface the HTTP status

RahulGopathi:
## Problem

When the offer endpoint returns a 4xx (e.g. 429 quota exceeded, 401/403), `negotiate()` swallows the rejection at debug level and re-sends an identical offer every 2s until `maxReconnectionAttempts`, then rejects `connect()` with a statusless `TransportStartError`.

Meanwhile ICE gathering — started by `setLocalDescription` — trickles candidates via `PATCH` with `pc_id: null` for a peer connection the server never created, producing a stream of 422s alongside the 4xx.

The consumer can't distinguish "server refused" from "network flake" and often stacks its own reconnect loop on top.

## Fix

- 4xx from the offer → reject `connect()` with an `RTVIError` carrying the HTTP status, tear down, no retry. 5xx / network failures keep the existing retry behaviour.
- `_canSendIceCandidates` is only enabled once an answer has been received, so a failed offer (refused or retrying) never PATCHes `pc_id: null`.
- `attemptReconnection()` honours the abort signal so `disconnect()` cancels the 2s/5s timers instead of resurrecting a peer connection after teardown.

## Tests

New `tests/src/transports/smallWebRtcOfferRefusal.spec.ts` (stubbed `RTCPeerConnection` + mocked `makeRequest`), 4 cases:
- 4xx rejects `connect()` with the status and the offer is not re-sent
- 4xx sends no trickle-ICE PATCH
- 5xx still retries, but ICE candidates wait for an answer
- `disconnect()` cancels a pending re-offer

All 4 fail on `main` and pass with this change. Full suite 80/80, prettier clean, `parcel build` OK.

Note: `npm run lint` fails on `main` as well (ESLint 9 with no `eslint.config.*` in the repo) — untouched here.

### DIFF
--- a/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
+++ b/transports/small-webrtc-transport/src/smallWebRTCTransport.ts
@@ -86,6 +86,13 @@ class SignallingMessageObject {
   }
 }
 
+// A 4xx means the server refused this offer; re-sending it can't succeed.
+const isOfferRefused = (e: unknown): e is Response =>
+  typeof Response !== "undefined" &&
+  e instanceof Response &&
+  e.status >= 400 &&
+  e.status < 500;
+
 const AUDIO_TRANSCEIVER_INDEX = 0;
 const VIDEO_TRANSCEIVER_INDEX = 1;
 const SCREEN_VIDEO_TRANSCEIVER_INDEX = 2;
@@ -566,6 +573,10 @@ export class SmallWebRTCTransport extends Transport {
   private async attemptReconnection(
     recreatePeerConnection: boolean = false
   ): Promise<void> {
+    if (this._abortController?.signal.aborted) {
+      logger.debug("Transport disconnected, skipping reconnection.");
+      return;
+    }
     if (this.isReconnecting) {
       logger.debug("Reconnection already in progress, skipping.");
       return;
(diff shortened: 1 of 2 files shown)


markbackman:
cc @filipi87 to take a look.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
filipi87, line 1: Tagging @mattieruth to review here, since she is working in similar fixes in this other PR:
- https://github.com/pipecat-ai/pipecat-client-web-transports/pull/178

mattieruth: oh thank you! I totally missed this PR. i'll see how i can combine

mattieruth:
Thanks @RahulGopathi -- apologies i had overlooked this when starting my own fixes. #178 is now rebased on top of these changes and folds our fixes together. closing this PR in lieu of that one.

RahulGopathi:
> Thanks @RahulGopathi -- apologies i had overlooked this when starting my own fixes. #178 is now rebased on top of these changes and folds our fixes together. closing this PR in lieu of that one.

no worries. Thanks

PR #175, opened 2026-08-27
docs: add the update-docs profile for the web transports

markbackman:
Third of the profiles that let this repo use the shared `update-docs` automation. Companions: [pipecat-cloud#201](https://github.com/daily-co/pipecat-cloud/pull/201) (merged), [pipecat-client-web#247](https://github.com/pipecat-ai/pipecat-client-web/pull/247).

## Ownership comes first in this one

These pages sit in the same docs directory as one owned by `pipecat-client-web`, and the split is by file rather than directory:

| Page | Owner |
|---|---|
| `js/transports/transport.mdx` | **pipecat-client-web** — the abstract `Transport` base |
| the six concrete transport pages | this repo |

The two profiles are complementary by construction: this one claims exactly the six the other disclaims, and names `transport.mdx` only to say never to edit it from here. A change here that appears to need it means a transport has diverged from the base it implements — that's a cross-repo finding, not an edit.

## Three things the tables encode that aren't derivable

**Directory names don't match page names.** `gemini-live-websocket-transport` → `gemini.mdx`, `openai-realtime-webrtc-transport` → `openai-webrtc.mdx`. So the pattern rule is an explicit table, not a transformation an agent could infer and get subtly wrong.

**`directToLLMBaseWebSocketTransport.ts` is a base class, not a skip.** It isn't re-exported from its package root, so the obvious call is to skip it. But it defines `LLMServiceOptions`, which `GeminiLLMServiceOptions` extends and which `gemini.mdx` documents field by field — its surface is observable even though the class isn't. This is exactly the case the contract's skip-list test is for: *can someone change or observe this without subclassing it?*

**A new transport needs two registrations.** `docs.json`, and the comparison table in `client/concepts/choosing-a-transport.mdx`. A transport absent from that page is one nobody picks, since it's where a reader arrives before knowing the transport exists.

## The defaults rule has a worked example

The section vocabulary says a `<ParamField>` default must be the value in source, not the value the docs wish were true — and points at MoQ's `clientId`/`botId` as the case. The client defaults to `client0`/`bot0` while a Pipecat bot publishes under `response` and listens on `request`, so the two never meet on defaults alone. The page documents the real defaults and warns about the mismatch rather than quietly printing values that would work.

That mismatch is also on the list of upstream bugs worth fixing at the source, separately from documenting it.

## Contract checks

- **Forward:** all 17 in-scope source files reach a rule
- **Backwards:** all 6 owned pages are reachable by one
- **Cross-profile:** the only page named by both profiles is `transport.mdx`, and this one names it to disclaim it

## Not included: the workflow

Deliberate, as with the other two. `update-docs.yml` should copy whatever the end-to-end dispatch test proves out on pipecat rather than being written speculatively.

### DIFF
--- a/.claude/skills/update-docs/SOURCE_DOC_MAPPING.md
+++ b/.claude/skills/update-docs/SOURCE_DOC_MAPPING.md
@@ -1,0 +1,178 @@
+# Source-to-Doc Mapping — pipecat-client-web-transports
+
+The profile for the shared `update-docs` skill, which lives in
+`pipecat-ai/pipecat` at `.claude/skills/update-docs/SKILL.md` and is published
+through the `pipecat-dev-skills` marketplace. `PROFILE_CONTRACT.md` beside it
+describes what this file has to provide.
+
+## Ownership — read this first
+
+The pages this repo documents sit in the same directory as pages owned by
+`pipecat-ai/pipecat-client-web`. The split is by file, not by directory:
+
+| Page | Owner |
+| --- | --- |
+| `api-reference/client/js/transports/transport.mdx` | **pipecat-client-web** — the abstract `Transport` base class |
+| `api-reference/client/js/transports/daily.mdx` | this repo |
+| `api-reference/client/js/transports/gemini.mdx` | this repo |
+| `api-reference/client/js/transports/moq.mdx` | this repo |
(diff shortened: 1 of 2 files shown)


mattieruth (approved):
Image

PR #176, opened 2026-08-28
ci(docs): open documentation PRs when public API changes

markbackman:
Installs the docs automation this repo has never had. **Stacked on #175** (the profile). This completes the rollout across all four repos.

## Not a rewrite

pipecat's workflow with four adaptations, so its accumulated hardening carries over: the **90-turn cap** that fixed a ~40% silent-drop rate, **formatting before `llms.txt` generation** (the reverse order fails the docs repo's lint), **outcome recording**, and **failure reporting** back onto the source PR.

| | |
|---|---|
| **Scope** | `transports/*/src/**`, excluding spec and test files |
| **Branch names** | `docs/transports-pr-N` |
| **Token scope** | App token for `owner: pipecat-ai`, scoped to `docs` alone |
| **Source repo refs** | every `gh` call targets this repo |

## Ownership note in the header

This repo owns the **six concrete transport pages** under `api-reference/client/js/transports/`. The seventh, `transport.mdx`, is the abstract base class and belongs to `pipecat-client-web`. The profile says so, but the ownership split is the thing most likely to be got wrong by someone reading only the workflow, so it's stated here too.

## Secrets are already set

`DOCS_BOT_APP_ID`, `DOCS_BOT_PRIVATE_KEY`, and `ANTHROPIC_API_KEY` are provisioned. The App credentials were verified directly against GitHub — it authenticates as `pipecat-docs-bot` (id 4080679) with `contents: write` and `pull_requests: write` on `pipecat-ai/docs`. It needs no grant on this repo, since the workflow reads its own PRs with the automatic `GITHUB_TOKEN`.

## Honest caveat

The shared skill has been verified statically — all 17 in-scope source files reach a rule, all 6 owned pages are reachable, the YAML parses with 12 steps — but has **not been run end to end** since being refactored. The first merged PR here will be its first live exercise, and a failure surfaces as a labelled red run rather than a wrong docs PR.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -180,7 +180,7 @@ jobs:
             GH_TOKEN=$DOCS_TOKEN gh pr create \
               --repo pipecat-ai/docs \
               --label auto-docs \
-              --label pipecat \
+              --label pipecat-client-web \
               --title "docs: update for pipecat PR #${{ steps.pr.outputs.number }}" \
               --body "$(cat <<'BODY'
             Automated documentation update for [transports PR #${{ steps.pr.outputs.number }}](https://github.com/pipecat-ai/pipecat-client-web-transports/pull/${{ steps.pr.outputs.number }}).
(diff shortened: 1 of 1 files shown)


mattieruth (approved):
Image

PR #177, opened 2026-09-02
fix(small-webrtc-transport): ignore signalingstatechange events from stale peer connections

### LINKED ISSUE #169: Reconnect loop never ends: attempt counter gets reset by the old peer connection
If the network goes away mid-session, the transport retries the offer every 2s and keeps doing that forever. It never hits `maxReconnectionAttempts`, so `stop()` never runs and `onDisconnected()` never fires. Our app had no idea the session was dead, the UI just sat there looking connected. `this.state` also stays `"ready"` the whole time (it's only set in `_connect`/`_disconnect`/`initDevices`), so there's nothing else to watch either.

The mic stays open through all of this too, so the browser keeps showing the recording indicator long after the call is gone.

As far as I can tell the cause is here:

```ts
// transports/small-webrtc-transport/src/smallWebRTCTransport.ts:497
pc.addEventListener("signalingstatechange", () => {
  logger.debug(`signalingState: ${this.pc!.signalingState}`);
  if (this.pc!.signalingState == "stable") {
    this.handleReconnectionCompleted();
  }
});
```

The listener is on `pc`, but it checks `this.pc`. Every retry does:

- `attemptReconnection(true)` bumps `reconnectionAttempts`
- `startNewPeerConnection()` sets `this.pc` to a brand new connection, which starts out `"stable"`
- `negotiate()` fails, gets caught, schedules the next attempt in 2s
- `closePeerConnection(oldPC)` runs and the old connection fires `signalingstatechange`

That last event is from the old connection but the handler looks at the new one, sees `"stable"`, and calls `handleReconnectionCompleted()`, which zeroes the counter. So it goes 1, 0, 1, 0 and never reaches 3.

To reproduce: connect to a bot, wait until it's talking, then turn off Wi-Fi (DevTools offline mode doesn't work, it doesn't affect WebRTC). Expected the transport to give up after 3 tries and call `stop()`. Instead `POST /offer` fails every 2s indefinitely. Seeing more than 3 attempts is enough to show the counter is being reset, or you can poll `transport.reconnectionAttempts` and watch it bounce between 1 and 0.

Fix is probably just ignoring events from connections that aren't current:

```ts
pc.addEventListener("signalingstatechange", () => {
  if (pc !== this.pc) return;
  logger.debug(`signalingState: ${pc.signalingState}`);
  if (pc.signalingState == "stable") {
    this.handleReconnectionCompleted();
  }
});
```

Using `pc` instead of `this.pc` also fixes the debug line, which right now logs the state of a different connection than the one that fired.

Happy to open a PR if that looks right.

mattieruth:
**NOTE**: I've rebased this on top of #178 to take advantage of the new test infrastructure it introduces. We should merge that one first.

## Summary
- The `signalingstatechange` listener was attached to a specific `pc` instance but checked `this.pc` (the current, possibly-reassigned connection) instead of `pc` itself.
- During reconnection, `startNewPeerConnection()` reassigns `this.pc` to a freshly created connection (which starts in the `"stable"` state) before the old connection is closed. Closing the old connection fires `signalingstatechange` on it, and the handler reads `this.pc` (the new, unrelated connection), sees `"stable"`, and calls `handleReconnectionCompleted()`, resetting `reconnectionAttempts` to 0.
- The counter therefore oscillates between 1 and 0 and never reaches `maxReconnectionAttempts`, so `stop()`/`onDisconnected()` never fire and the mic indicator stays on indefinitely after the network drops.
- Fix: bail out early if the event did not come from the currently active connection (`pc !== this.pc`), and read state off `pc` (the connection that actually fired the event) rather than `this.pc`.

Fixes #169

## Test plan
- [ ] Connect to a bot, wait until audio is flowing, then disable Wi-Fi (DevTools offline mode does not affect WebRTC) and confirm the transport gives up after `maxReconnectionAttempts` and calls `stop()`/`onDisconnected()`
- [ ] Confirm `transport.reconnectionAttempts` increments monotonically during a real outage instead of bouncing between 0 and 1
- [ ] Confirm mic indicator turns off once `stop()` runs

### DIFF
--- a/tests/src/helpers/fakeRTCPeerConnection.ts
+++ b/tests/src/helpers/fakeRTCPeerConnection.ts
@@ -26,19 +26,46 @@ export class FakeRTCPeerConnection {
   sctp: { maxMessageSize: number } | null = null;
   onicecandidate: ((event: { candidate: unknown }) => void) | null = null;
 
+  // Real per-instance listener registry (rather than a bare vi.fn() no-op)
+  // so tests can simulate the browser actually firing events — e.g. close()
+  // firing "signalingstatechange" on itself, which is what a stale peer
+  // connection does mid-reconnection.
+  private listeners = new Map<string, Set<() => void>>();
+
   createOffer = vi.fn(async () => ({ type: "offer", sdp: "v=0\r\n" }));
   setLocalDescription = vi.fn(async (desc: RTCSessionDescriptionInit) => {
     this.localDescription = desc;
   });
-  setRemoteDescription = vi.fn(async () => {});
+  setRemoteDescription = vi.fn(async () => {
+    // Real browsers land in "stable" once a valid answer is applied, and
+    // fire signalingstatechange as part of that transition.
+    this.signalingState = "stable";
+    this.dispatch("signalingstatechange");
+  });
   addTransceiver = vi.fn();
-  addEventListener = vi.fn();
-  removeEventListener = vi.fn();
   getTransceivers = vi.fn(() => []);
   getSenders = vi.fn(() => []);
-  close = vi.fn();
-
   createDataChannel = vi.fn(() => new FakeRTCDataChannel());
+
+  addEventListener = vi.fn((type: string, cb: () => void) => {
+    if (!this.listeners.has(type)) this.listeners.set(type, new Set());
+    this.listeners.get(type)!.add(cb);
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
Makes sense. Nice catch! 🚀

PR #178, opened 2026-09-02
fix(small-webrtc-transport): don't swallow HTTP errors from the offer endpoint

### LINKED ISSUE #173: SmallWebRTCTransport swallows HTTP errors from the offer endpoint (blind retries, statusless final rejection)
## Description

When the `/offer` endpoint refuses the SDP offer with an HTTP error (in our case a `429 Too Many Requests` from a usage-limit check), `SmallWebRTCTransport` swallows the error entirely. The caller of `PipecatClient.connect()` never sees the HTTP status — or any error object describing what happened — and the transport produces a multi-request retry storm before failing with a generic, statusless rejection.

Observed on `@pipecat-ai/small-webrtc-transport` 1.10.5; the relevant code is unchanged on current `main`.

## What happens

1. `negotiate()` POSTs the offer, the server responds 429. The rejection is caught here and only logged at **debug** level, then a blind retry is scheduled:

   https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/small-webrtc-transport/src/smallWebRTCTransport.ts#L779-L784

   ```ts
   } catch (e) {
     logger.debug(`Reconnection attempt ${this.reconnectionAttempts} failed: ${e}`);
     this.isReconnecting = false;
     setTimeout(() => this.attemptReconnection(true), 2000);
   }
   ```

   A 4xx from the signaling server is deterministic — retrying it every 2s cannot succeed, but the transport retries up to the reconnection cap anyway.

2. Meanwhile the connection flow continues past the failed offer, so queued ICE candidates get flushed to the server with `pc_id: null` (the answer that would have carried `pc_id` never arrived). Against a server that validates the PATCH body, every one of these is a 422.

3. After `attemptReconnection` (https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/small-webrtc-transport/src/smallWebRTCTransport.ts#L566) exhausts its attempts, `stop()` rejects the pending connect promise with a bare `TransportStartError`:

   https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/small-webrtc-transport/src/smallWebRTCTransport.ts#L977-L978

   No message, no status, no cause — the original `Response`/error from step 1 is gone.

## Why it matters

The application cannot distinguish "the server refused this session (quota, auth, policy)" from "the network dropped". Both surface identically, so any app-level reconnect logic re-runs the whole storm. In our app one orb click while over quota produced ~20–40 requests (transport retries × app retries: offer POSTs + `pc_id: null` PATCHes) and a UI stuck on "Connecting" — versus the one 429 that should have ended it.

We've worked around it by moving authorization to a pre-negotiation endpoint via `startBotAndConnect()` (whose `startBot` correctly preserves the status in `StartBotError`), but an offer that fails *after* a successful start still hits this same swallow path.

## Expected behavior

Some combination of:

1. **Don't retry deterministic failures**: if the offer request rejects with a `Response` whose status is 4xx, fail fast instead of scheduling `attemptReconnection`.
2. **Preserve the cause on final failure**: when giving up, reject the connect promise with an error carrying the original status/message (e.g. `TransportStartError` with the status, or `cause` set to the original error) instead of a bare `TransportStartError`.
3. **Don't flush ICE candidates after a failed offer** — the `pc_id: null` PATCHes are pure noise against a validating server.

## Repro

Any signaling server that answers `POST /offer` with a 4xx:

```ts
const client = new PipecatClient({ transport: new SmallWebRTCTransport({ offerUrlTemplate: ... }), ... });
await client.connect({ webrtcRequestParams: { endpoint: "https://host/offer-that-returns-429" } });
```

Watch the network tab: 1 + N offer POSTs 2s apart, interleaved ICE PATCHes with `pc_id: null`, and a final rejection that contains no trace of the 429.

mattieruth:
## Summary
- `negotiate()` caught every offer-request failure the same way: log at debug level and blindly retry on a 2s timer, even for a deterministic 4xx (e.g. a 429 from a usage-limit check) that cannot possibly succeed by retrying.
- The final rejection was a bare, traceless `TransportStartError` — no status, no message, no reference to the original error.
- ICE candidates kept flushing on their own 200ms timer regardless of whether the offer ever succeeded, so a failed offer still produced PATCH requests with a null `pc_id` against a server that never sent an answer.

Changes:
- Fail fast on a 4xx response from the offer endpoint instead of scheduling another `attemptReconnection`.
- Preserve the original status/error on the final rejection: `TransportStartError`'s message now includes the status (or stringified error), and the original error is attached via `.cause`. Applied both to the new fast-fail path and to the existing "max reconnection attempts exhausted" path.
- Only start flushing queued ICE candidates once `negotiate()` gets a successful answer, instead of unconditionally after every negotiation attempt.

Fixes #173

## Test plan
- [x] Added `tests/src/transports/smallWebRtcOfferFailure.spec.ts` (4 new tests, driven through the public `transport.connect()` API with fake timers + a mocked global `fetch`):
  - a 429 offer response fails fast (1 fetch call total, no retry after 5s) and the rejection carries the status + `cause`
  - a non-`Response` failure (e.g. a network `TypeError`) still retries through all `maxReconnectionAttempts`, and the final rejection carries the last error as `cause`
  - a failed offer with an ICE candidate queued concurrently never produces a PATCH
  - a successful offer/answer allows a subsequently queued ICE candidate to be flushed
- [x] Full existing suite (`vitest run`) still passes: 9 files, 80 tests
- [x] `tsc --noEmit` clean on both `transports/small-webrtc-transport` and `tests`


devansh-sodhi:
Damn can i get some credit

mattieruth:
> Damn can i get some credit

Thanks again for the thorough write-up. It made tracking this down straightforward. In open source, credit generally follows the PR rather than the issue report, so next time feel free to open a PR with your proposed fix if you’d like your name on the commit.

devansh-sodhi:
Damn mattie , all that credit goes to claude tbf neither of us deserve it

tests/src/transports/smallWebRtcOfferFailure.spec.ts
filipi87, line 1: Hi @mattieruth, this [PR](https://github.com/pipecat-ai/pipecat-client-web-transports/pull/174/changes) was also handling this exact same case. Maybe it’s worth taking a look to see if there’s anything there we should reuse here.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -775,7 +776,32 @@ export class SmallWebRTCTransport extends Transport {
...
+          `Offer request failed with non-retryable status ${status}. Giving up.`
+        );
+        this.isReconnecting = false;
+        // Deferred (rather than awaited inline) so that on the very first
+        // connection attempt, _connect() has a chance to finish wiring up
+        // _connectFailed before stop() tries to call it.
+        const error = this.buildTransportStartError();
+        setTimeout(() => void this.stop(error), 0);
+        return;
+      }
```
filipi87, line 815: In this case, shouldn’t we throw an error so that, for example, inside `_connect`, we could catch it and change the transport state to error (`this.state = "error";`) ?

Otherwise, wouldn’t this mean that `onConnected` would still be invoked before we call `stop` here, since we’re deferring the `stop` call?

mattieruth: fixing the weird deferred stop call. but in general, stop() with an error will call this._connectFailed() which is the rejection/throw that will bubble up. But good call on the this.state = 'error' -- unrelated to these changes, nothing was setting that. that said, in the client.connect() handling, if client._connect() throws an error, it immediately calls disconnect() which will change the state to 'disconnecting' -> 'disconnected' so that 'error' state is very fleeting.

transports/small-webrtc-transport/src/smallWebRTCTransport.ts
```diff
@@ -86,6 +85,13 @@ class SignallingMessageObject {
   }
 }
 
+// A 4xx means the server refused this offer; re-sending it can't succeed.
```
filipi87, line 88: Nice.

filipi87 (approved):
LGTM! Nice improvement. 🚀

devansh-sodhi:
any idea When will pipecat release this change?

mattieruth:
If I don't get to it today, probably Tuesday of next week.

PR #180, opened 2026-09-02
fix(small-webrtc-transport): guard keepalive ping against InvalidStateError in Safari

### LINKED ISSUE #63: SmallWebRTCTransport: InvalidStateError: The object is in an invalid state.
Hello, we recently experienced the following error that is happening only in Safari:

Image

The screenshot comes from the internal chat bot and we can reproduce the issue when we allow the chat bot to say first message, we will respond and after his second message close the connection. I tracked down the issue to be in this part:

```javascript
this.keepAliveInterval = setInterval(() => {
  const message = "ping: " + new Date().getTime();
  dc.send(message);
}, 1000);
```

I did a quick fix that solved the issue on our side:

```javascript
const originalSend = dc.send.bind(dc) as (data: string | ArrayBuffer | ArrayBufferView | Blob) => void;

dc.send = (message: string | ArrayBuffer | ArrayBufferView | Blob) => {
  if (dc.readyState === 'open') {
    return originalSend(message);
  }
  // Silently fail for non-open states (closing, closed, connecting)
  return false;
};
```

However, I was unsure if this is the right way to go so I decided to reach out and ask you for help.

mattieruth:
**NOTE: I've rebased this on top of https://github.com/pipecat-ai/pipecat-client-web-transports/pull/177 and #178 to take advantage of the new test infrastructure they introduce. We should merge those first.**

## Summary
- `RTCDataChannel` transitions through `"open" -> "closing" -> "closed"`, but there's no `"closing"` event — only `"close"`, which is what actually clears the keepalive interval (`createDataChannel()`'s `dc.addEventListener("close", ...)` handler).
- If the 1s ping tick lands while the channel is `"closing"` but not yet `"closed"`, `dc.send()` gets called on a non-open channel. Chrome tolerates this; Safari throws `InvalidStateError: The object is in an invalid state.`, matching the report.
- Fix: guard the ping on `dc.readyState === "open"` instead of sending unconditionally. Narrower than the reporter's suggested fix (monkey-patching `dc.send` globally) since every other call site in the file (`sendMessage`, `sendSignallingMessage`, `syncTrackStatus`) already guards `readyState` before calling `dc.send` — only the keepalive interval was missing the check.

Fixes #63

## Test plan
- [x] Added `tests/src/transports/smallWebRtcKeepAlive.spec.ts`: drives a successful connect, opens the fake data channel, confirms the keepalive ping sends while `readyState === "open"`, then flips `readyState` to `"closing"` (simulating the real-world race) and confirms the next tick is skipped rather than calling `send()`.
- [x] Verified the test fails (extra `send()` call) against the pre-fix interval and passes with the fix.
- [x] Full suite passes: 11 files, 83 tests.
- [x] `tsc --noEmit` clean on both `transports/small-webrtc-transport` and `tests` (two pre-existing, unrelated TypedArray errors in `dailyMediaManager.ts`/`mediaManager.ts` aside).

Branch is based on `mb/fix-reconnect-loop-issue-169` (#177), so this PR depends on #177 (and transitively #178) merging first — the diff above is just this fix's incremental change.

### DIFF
--- a/tests/src/helpers/fakeRTCPeerConnection.ts
+++ b/tests/src/helpers/fakeRTCPeerConnection.ts
@@ -14,8 +14,24 @@ export class FakeRTCDataChannel {
   readyState: string = "connecting";
   send = vi.fn();
   close = vi.fn();
-  addEventListener = vi.fn();
-  removeEventListener = vi.fn();
+
+  // Real per-instance listener registry so tests can simulate the browser
+  // actually firing "open" / "close" — e.g. to drive the keepalive-interval
+  // setup/teardown in createDataChannel().
+  private listeners = new Map<string, Set<(ev?: unknown) => void>>();
+
+  addEventListener = vi.fn((type: string, cb: (ev?: unknown) => void) => {
+    if (!this.listeners.has(type)) this.listeners.set(type, new Set());
+    this.listeners.get(type)!.add(cb);
+  });
+  removeEventListener = vi.fn((type: string, cb: (ev?: unknown) => void) => {
+    this.listeners.get(type)?.delete(cb);
+  });
+
+  /** Simulates the browser firing `type` on this channel. */
+  dispatch(type: string, ev?: unknown) {
+    this.listeners.get(type)?.forEach((cb) => cb(ev));
+  }
 }
 
 export class FakeRTCPeerConnection {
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
LGTM!

PR #191, opened 2026-09-10
fix(moq-transport): hold client-ready until the bot audio subscription is live

vipyne:
TL;DR: this makes `client-ready` event fire at correct time for MoQ transport, so beginning of first message is not cut off.

## What

  Delays sending the RTVI `client-ready` message until the transport's
  subscription to the bot's audio track has been issued.

  `sendReadyMessage()` is now async (the SDK already awaits it): it waits for
  the `@moq/watch` side to reach the state where the audio-track SUBSCRIBE is
  on the wire, then flips to `ready` and sends `client-ready`. A 10s timeout
  falls back to sending immediately with a console warning, so `connect()`
  can't hang against a bot that publishes no audio track.

  ## Why

  `client-ready` is the bot's cue to start speaking, and the bot's audio is
  live media with no replay — anything published before our subscription
  exists is lost.

  Today `client-ready` is sent as soon as `connect()` returns, and because
  the client→bot transcript track replays its log to a late-subscribing bot,
  the message reaches the bot the moment it comes up on a cold start. The
  audio subscription, by contrast, needs several round trips first:
  announcement → catalog subscribe → catalog → rendition selection →
  audio-track SUBSCRIBE. The bot's greeting lands in that gap and the head
  of its first utterance is cut off.

  This was flagged in pipecat-ai/pipecat#5484 (see
  https://github.com/pipecat-ai/pipecat/pull/5484#discussion_r3973148049):
  `on_client_ready` should mean the client is subscribed and ready to
  receive audio, so the fix belongs here rather than only bot-side.

  ## How

  `_waitForBotAudio()` watches `source.track`, `source.config`, and
  `broadcast.active` — exactly the signals that gate `Watch.Audio.Decoder`'s
  internal subscribe effect — and resolves one macrotask after all three are
  set, so the decoder's SUBSCRIBE goes out first. It deliberately does not
  wait for received audio bytes: the bot won't speak until `client-ready`
  arrives, so that would deadlock.

  A small race remains (the SUBSCRIBE propagating relay→bot vs.
  `client-ready` arriving on the transcript track — one relay hop); the
  bot-side `audio_out_subscriber_timeout` from pipecat-ai/pipecat#5484
  still covers that window and non-SDK clients. The two fixes are
  complementary.

### DIFF
--- a/transports/moq-transport/src/moqTransport.ts
+++ b/transports/moq-transport/src/moqTransport.ts
@@ -688,6 +688,8 @@ export class MoqTransport extends Transport {
   private _waitForBotAudio(timeoutMs = 10_000): Promise<void> {
     const source = this._audioSource;
     const broadcast = this._watchBroadcast;
+    // Both are set in _connect(); if they're null
+    // (bc a `disconnect` happened); resolve immediately;
     if (!source || !broadcast) return Promise.resolve();
     return new Promise((resolve) => {
       let done = false;
--- a/transports/moq-transport/src/moqTransport.ts
+++ b/transports/moq-transport/src/moqTransport.ts
@@ -671,13 +671,52 @@ export class MoqTransport extends Transport {
     }
   }
 
-  sendReadyMessage(): void {
-    // Flip transport state to `ready` and dispatch the RTVI `client-ready`
-    // message on the client→bot transcript track so the bot's RTVIProcessor
-    // can negotiate protocol version. Mirrors the pattern used by
-    // small-webrtc-transport and daily-transport.
+  async sendReadyMessage(): Promise<void> {
+    // `client-ready` is the bot's cue to start speaking, and its audio is
+    // live media with no replay — so hold it until our audio subscription
+    // is on the wire, or the head of the first utterance is lost.
+    await this._waitForBotAudio();
     this.state = "ready";
     this.sendMessage(RTVIMessage.clientReady());
+  }
+
+  /** Resolve once the watch side has issued its audio-track subscribe —
+   *  `Watch.Audio.Decoder` subscribes in an effect gated on exactly these
+   *  three signals. Not gated on received audio: the bot won't speak
+   *  until `client-ready` arrives. The timeout keeps `connect()` from
+   *  hanging against a bot with no audio track. */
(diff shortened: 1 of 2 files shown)


transports/moq-transport/src/moqTransport.ts
```diff
@@ -671,15 +671,54 @@ export class MoqTransport extends Transport {
...
 
+  /** Resolve once the watch side has issued its audio-track subscribe —
+   *  `Watch.Audio.Decoder` subscribes in an effect gated on exactly these
+   *  three signals. Not gated on received audio: the bot won't speak
+   *  until `client-ready` arrives. The timeout keeps `connect()` from
+   *  hanging against a bot with no audio track. */
+  private _waitForBotAudio(timeoutMs = 10_000): Promise<void> {
+    const source = this._audioSource;
+    const broadcast = this._watchBroadcast;
+    if (!source || !broadcast) return Promise.resolve();
```
filipi87, line 693: If either doesn't exist, we don't wait here. Is this the expected behavior? Should we expect both to always exist at this point?

transports/moq-transport/src/moqTransport.ts
filipi87, line 1: Just one question: can we be sure that this._audioSource and this._watchBroadcast will always be available at that point? Other than that, it all makes sense. 🚀

transports/moq-transport/src/moqTransport.ts
vipyne: unless a `disconnect` happens between connect and ready, source and broadcast will exist ([they](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/moq-transport/src/moqTransport.ts#L527) are both [assigned](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/moq-transport/src/moqTransport.ts#L549) in [`_connect`](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/moq-transport/src/moqTransport.ts#L361).

vipyne: and if a disconnect does happen, we do want to resolve immediately.

filipi87 (approved):
LGTM! 🚀

PR #192, opened 2026-09-11
fix: Fixed issue with mute toggle states being delayed

mattieruth:
Fixed an issue in the Daily Transport and MediaManager where the mic/camEnabled flag was stale in the track-started/stopped events. Previously, the flag was based on daily's localVideo()/Audio(), but those have not yet updated at the moment track-started/stopped are called, so if you rely on those in your handler, the state is wrong.


markbackman (approved):
🙌

PR #213, opened 2026-09-16
feat(moq-transport): number transcript records and drop replays on resubscribe

mattshep:
The MoQ transcript tracks are single-group JSON streams that a subscriber always reads from the first record, so a reconnect on either side replays the whole log and the bot receives `client-ready` again. This numbers what the transport sends and drops what it has already received, matching the bot side of the same format in pipecat-ai/pipecat#5800.

- **Outbound records.** `sendMessage` appends the RTVI message plus `seq`, its position in the connection's log, and `epoch`, an id minted per connection. The replay a late subscriber gets carries the same numbers, so the bot can skip it.
- **Inbound records.** `_drainTranscript` keeps the last `seq` accepted for the bot's current `epoch`, drops anything at or below it, and strips both fields before `PipecatClient` sees the message. A new `epoch` is a new bot and starts the count over. Records without `seq` are delivered unchanged, so bots that predate the fields keep working, and the bot's `session-ending` marker is recognised either way.
- **Exports.** `acceptTranscriptRecord` and `TranscriptRecord` are exported for other implementations of the stream, and the README documents the record format.

- **Session-ending marker.** `disconnect()` appends `{ label: "moq-transport", type: "session-ending" }` to the client's transcript before tearing down, so the bot ends the call at once. When the bot's tracks end without its own marker, which is also what a failed relay between the peers looks like, the transport redials through `@moq/net`: the bot either reappears on the new session or is never announced on it.


vipyne:
we should consider [mattieruth's comment](https://github.com/pipecat-ai/pipecat-client-web-transports/pull/215#pullrequestreview-5250289128) - is it heavy-handed to change state when a transport temporarily drops?  I agree we don't want to introduce too much variance between transports.

mattshep:
Agreed, and I've changed it here in f25ebc1 rather than leave it for later. The status mirroring that flips the state during a reconnect is on main already, so this PR was carrying it either way.

Once the session is ready, the relay status no longer drives the transport state. `Connection.Reload` redials in the background and the state stays on `ready`, the same as the other transports through their own reconnects. That matters more than it looks: `PipecatClient` only lets `sendText`, `sendClientMessage`, `appendToContext` and the rest through while the state is exactly `ready`, so every relay blip was a window where app code got `BotNotReadyError`.

Holding `ready` is also accurate for this transport, which is what made me comfortable with it. A message sent during the gap is kept in the transcript log, the log is replayed when the bot resubscribes, and the sequence numbers this PR adds let the bot drop what it already has. Audio from the gap is not replayed, and the README says so.

I checked it against a real failure: a session at `ready`, then the relay killed and restarted under it. The relay connection dropped and redialed, the transport reported no state change at all, a `sendClientMessage` issued while the connection was down did not throw, and that message was in the log served by the new relay process afterwards.

Two things had to change with it, because holding `ready` would otherwise have made them worse:

- **Reload giving up.** When Reload exhausts its retry window it rejects `closed`, which nothing was listening for, so the state sat on `connecting` for good and the rejection went unhandled. The transport now moves to `error` and reports the dial failures as a fatal error, the way the Daily transport handles its fatal errors, and `PipecatClient` disconnects.
- **A relay that is never reached.** `sendReadyMessage` waits up to 10 s for the bot's audio subscribe and then goes to `ready` regardless, even if the relay never connected. On main the next failed retry knocked that back out; with `ready` held it would have stuck for the whole retry window. The timeout now starts once the relay is connected, and a disconnect settles a wait that is still pending instead of letting it land on `ready` afterwards. Checked in a browser against a dead port: no `ready` after 14 s, and a disconnect mid-wait ends on `disconnected`.

What this leaves out on purpose is any signal for an app that wants to show "reconnecting". No transport has one today, so that reads as an SDK-level addition across all of them rather than something to invent here.

mattieruth (commented):
i only focused on the transport state changes and even then i'm a little hazy on how this transport works, but i'm pretty sure it looks good. you have my approval!

Image

mattieruth (approved):

filipi87 (approved):
The changes makes sense. LGTM! 🚀

vipyne (approved):
👍

PR #214, opened 2026-09-17
add _wasReady flag to handle moq-relay reconnects

vipyne:
`clientReady` is what sets `connect` button to `disconnect` when a session is active and it is only emitted **_once_** per session. 

if moq transport encounters an interruption, the reconnection leaves the "connect" button in a "connecting..." state because it never 'hears' a second `clientReady` after the reconnect. it's still the same bot session, the transport just temporarily dropped.

`_wasReady` flag allows the button state to correctly go back to `disconnect`.


vipyne:
superseded by https://github.com/pipecat-ai/pipecat-client-web-transports/pull/215

PR #215, opened 2026-09-17
add _wasReady flag to handle moq-relay reconnects

vipyne:


mattshep (approved):
The state fix is right. `sendReadyMessage` runs once, so without this the transport sits on `connected` after a reconnect and never returns to ready. Leaving `client-ready` unsent on the reconnect is right too: that message is the bot's cue to start speaking, and re-sending it would make the bot redo its greeting after a blip.

**Finding 1 — CI is red on `lint-commits`.** The commit subject carries no conventional-commit type, so commitlint rejects it. `fix: add _wasReady flag to handle moq-relay reconnects` clears it.

**Finding 2 — the reconnect path declares ready without re-waiting for the audio subscription.** `sendReadyMessage` holds the ready transition behind `_waitForBotAudio()` because the bot's audio is live media with no replay. The reconnect transition sets ready straight off the Reload status, so the state flips before the watch side has re-issued its subscribe, and a reconnect where the bot's broadcast does not come back reports ready with nothing on the wire. Awaiting `_waitForBotAudio()` on that path costs nothing in the normal case, and its 10s timeout resolves either way so it cannot wedge the reconnect.

**Finding 3 — typo.** "Allows the UI state to behavior correctly" reads as "behave".

One boundary worth recording while this is fresh: skipping `client-ready` on the reconnect assumes the peer on the other side of the gap is the same bot instance. That holds today because the bot's transport keeps its epoch across redials and stays subscribed, so it needs no second cue. A bot that came back as a fresh instance mid-session would wait for a `client-ready` this path no longer sends. Nothing to change here, but it is an assumption the relay work should not quietly break.

The lockfile changes look like a downgrade at a glance but they are correct: the workspace `package.json` files on the base branch already carry these versions, so the lockfile is catching up rather than moving anything. Worth a line in the PR body to save the next reader the check.

mattieruth (commented):
Sorry I'm late to the party, but I wanted to point out that this behavior does not match the other transports. In other transports, we don't change the transport state during reconnects at all. It remains 'ready' until you disconnect.

Maybe we should and we should discuss that, but at this time, it means that the state events for this transport are not consistent with other transports.

vipyne:
> Sorry I'm late to the party, but I wanted to point out that this behavior does not match the other transports. In other transports, we don't change the transport state during reconnects at all. It remains 'ready' until you disconnect.
> 
> Maybe we should and we should discuss that, but at this time, it means that the state events for this transport are not consistent with other transports.

following up in #213

